PR #1, opened 2024-05-30
Create Moddingway base

OTCompa:
This PR creates the base Moddingway bot and registers the basic moderation commands on Discord, which are only dummy functions at this point.

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -101,7 +101,7 @@ var UnmuteCommand = &discordgo.ApplicationCommand{
 		{
 			Type:        discordgo.ApplicationCommandOptionUser,
 			Name:        "user",
-			Description: "User being muted",
+			Description: "User being unmuted",
 			Required:    true,
 		},
 		{
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -80,7 +80,7 @@ var MuteCommand = &discordgo.ApplicationCommand{
 		{
 			Type:        discordgo.ApplicationCommandOptionString,
 			Name:        "duration",
-			Description: "Duration of mute (e.g \"1m\", \"1h\", \"1d\")",
+			Description: "Duration of mute (e.g \"1m, 1h, 1d\")",
 			Required:    true,
 		},
 		{
@@ -115,7 +115,7 @@ var UnmuteCommand = &discordgo.ApplicationCommand{
 var BanCommand = &discordgo.ApplicationCommand{
 	Name:                     "ban",
 	DefaultMemberPermissions: &adminPermission,
-	Description:              "Ban the specified user.",
+	Description:              "Ban the specified user and notify the user why via DMs.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
 			Type:        discordgo.ApplicationCommandOptionUser,
@@ -206,7 +206,7 @@ var SlowmodeCommand = &discordgo.ApplicationCommand{
 		{
 			Type:        discordgo.ApplicationCommandOptionString,
 			Name:        "duration",
-			Description: "Duration of slowmode (e.g \"1m\", \"1h\", \"1d\")",
+			Description: "Duration of slowmode (e.g \"1m, 1h, 1d\")",
 			Required:    true,
 		},
 	},
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -83,9 +83,9 @@ var MuteCommand = &discordgo.ApplicationCommand{
 			Required:    true,
 		},
 		{
-			Type:        discordgo.ApplicationCommandOptionInteger,
+			Type:        discordgo.ApplicationCommandOptionString,
 			Name:        "duration",
-			Description: "Duration in minutes",
+			Description: "Duration of mute (e.g \"1m\", \"1h\", \"1d\")",
 			Required:    true,
 		},
 		{
@@ -145,7 +145,7 @@ var UnbanCommand = &discordgo.ApplicationCommand{
 		{
 			Type:        discordgo.ApplicationCommandOptionUser,
 			Name:        "user",
-			Description: "User being unbanned",
+			Description: "User being unbanned (Discord ID)",
 			Required:    true,
 		},
 		{
@@ -209,9 +209,9 @@ var SlowmodeCommand = &discordgo.ApplicationCommand{
 	Description:              "Add slowmode to current channel.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
-			Type:        discordgo.ApplicationCommandOptionInteger,
+			Type:        discordgo.ApplicationCommandOptionString,
 			Name:        "duration",
-			Description: "Slowmode interval duration in seconds",
+			Description: "Duration of slowmode (e.g \"1m\", \"1h\", \"1d\")",
 			Required:    true,
 		},
 	},
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -8,12 +8,7 @@ import (
 
 // Set up vars for the DefaultMemberPermissions field in each command definition
 var (
-	// I am too dumb and tired to figure out how to set permission restrictions for the roles without these vars.. sorry..
-	kickPermission    int64 = discordgo.PermissionKickMembers
-	mutePermission    int64 = discordgo.PermissionModerateMembers
-	banPermission     int64 = discordgo.PermissionBanMembers
-	channelPermission int64 = discordgo.PermissionManageChannels
-	messagePermission int64 = discordgo.PermissionManageMessages
+	adminPermission int64 = discordgo.PermissionAdministrator
 )
 
 // AddCommands registers the slash commands with Discord
@@ -53,7 +48,7 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 
 var KickCommand = &discordgo.ApplicationCommand{
 	Name:                     "kick",
-	DefaultMemberPermissions: &kickPermission,
+	DefaultMemberPermissions: &adminPermission,
 	Description:              "Kick the specified user and notify the user why via DMs.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
@@ -73,7 +68,7 @@ var KickCommand = &discordgo.ApplicationCommand{
 
 var MuteCommand = &discordgo.ApplicationCommand{
 	Name:                     "mute",
-	DefaultMemberPermissions: &mutePermission,
+	DefaultMemberPermissions: &adminPermission,
 	Description:              "Mute the specified user.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
@@ -99,7 +94,7 @@ var MuteCommand = &discordgo.ApplicationCommand{
 
 var UnmuteCommand = &discordgo.ApplicationCommand{
 	Name:                     "unmute",
-	DefaultMemberPermissions: &mutePermission,
+	DefaultMemberPermissions: &adminPermission,
 	Description:              "Unmute the specified user.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
@@ -119,7 +114,7 @@ var UnmuteCommand = &discordgo.ApplicationCommand{
 
 var BanCommand = &discordgo.ApplicationCommand{
 	Name:                     "ban",
-	DefaultMemberPermissions: &banPermission,
+	DefaultMemberPermissions: &adminPermission,
 	Description:              "Ban the specified user.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
@@ -139,7 +134,7 @@ var BanCommand = &discordgo.ApplicationCommand{
 
 var UnbanCommand = &discordgo.ApplicationCommand{
 	Name:                     "unban",
-	DefaultMemberPermissions: &banPermission,
+	DefaultMemberPermissions: &adminPermission,
 	Description:              "Unban the specified user.",
 	Options: []*discordgo.ApplicationCommandOption{
 		{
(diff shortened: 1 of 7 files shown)


internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being banned",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for ban",
+			Required:    false,
```
brtran4, line 131: this should be true

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being kicked",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for kick",
+			Required:    false,
```
brtran4, line 65: should be true

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionInteger,
+			Name:        "duration",
+			Description: "Duration in minutes",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for mute",
+			Required:    false,
```
brtran4, line 91: change to true

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being muted",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for mute",
+			Required:    false,
```
brtran4, line 111: change to true

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User whose nickname to rename",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "nickname",
+			Description: "Nickname to rename user to",
+			Required:    true,
```
brtran4, line 185: i think we should add a required reason field

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+var RemoveNicknameCommand = &discordgo.ApplicationCommand{
+	Name:                     "removenickname",
+	DefaultMemberPermissions: &mutePermission,
+	Description:              "Remove the nickname of the specified user.",
+	Options: []*discordgo.ApplicationCommandOption{
+		{
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User whose nickname to rename",
+			Required:    true,
```
brtran4, line 165: same here, let's add a required reason field

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being unbanned",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for unban",
+			Required:    false,
```
brtran4, line 151: kinda on the fence here, but let's also make this true

internal/discord/handlers.go
brtran4, line 1: so if this file is only here for command helpers, we can rename this file to be command-handlers.go just to make it more specific and keep this file on the smaller side

brtran4 (changes requested):
even though this is a hassle, it would be good to do it now. let's add doc strings for every function explaining what each one does in a sentence or two.

we also want to send this to a specific channel that the mods will make in the future, so we'll prob want a channel ID variable somewhere

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+		{
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being muted",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionInteger,
+			Name:        "duration",
+			Description: "Duration in minutes",
```
clairefleuret, line 84: We should include hours/days

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+		{
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being muted",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionString,
+			Name:        "reason",
+			Description: "Reason for mute",
```
clairefleuret, line 110: should be unmute?

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+
+var SlowmodeCommand = &discordgo.ApplicationCommand{
+	Name:                     "slowmode",
+	DefaultMemberPermissions: &channelPermission,
+	Description:              "Add slowmode to current channel.",
+	Options: []*discordgo.ApplicationCommandOption{
+		{
+			Type:        discordgo.ApplicationCommandOptionInteger,
+			Name:        "duration",
+			Description: "Slowmode interval duration in seconds",
```
clairefleuret, line 198: should include minutes/hours

internal/discord/handlers.go
clairefleuret, line 65: should be required

internal/discord/handlers.go
clairefleuret, line 91: should be required

internal/discord/handlers.go
clairefleuret, line 111: should be required

internal/discord/handlers.go
clairefleuret, line 131: should be required

internal/discord/handlers.go
clairefleuret, line 151: should be required

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+
+var UnbanCommand = &discordgo.ApplicationCommand{
+	Name:                     "unban",
+	DefaultMemberPermissions: &banPermission,
+	Description:              "Unban the specified user.",
+	Options: []*discordgo.ApplicationCommandOption{
+		{
+			Type:        discordgo.ApplicationCommandOptionUser,
+			Name:        "user",
+			Description: "User being unbanned",
```
clairefleuret, line 144: might need to be more specific in unbans since the user is no longer in the server. Might need the actual discord ID

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+		{
+			Type:        discordgo.ApplicationCommandOptionChannel,
+			Name:        "channel",
+			Description: "Channel to purge",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionInteger,
+			Name:        "message-number",
+			Description: "Number of messages to purge",
```
clairefleuret, line 224: we might need to implement a max for this, im not sure how many resources are requred for a normal purge

internal/discord/handlers.go
brtran4: duplicate comment. closing

internal/discord/handlers.go
brtran4: duplicate comment. closing

internal/discord/handlers.go
brtran4: duplicate comment. closing

internal/discord/handlers.go
brtran4: resources shouldn't be an issue, it would take longer to purge more messages

internal/discord/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/bwmarrin/discordgo"
+)
+
+var (
+	// I am too dumb and tired to figure out how to set permission restrictions for the roles without these vars.. sorry..
+	kickPermission    int64 = discordgo.PermissionKickMembers
+	mutePermission    int64 = discordgo.PermissionModerateMembers
+	banPermission     int64 = discordgo.PermissionBanMembers
+	channelPermission int64 = discordgo.PermissionManageChannels
+	messagePermission int64 = discordgo.PermissionManageMessages
```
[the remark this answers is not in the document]
OTCompa: These vars are meant for limiting who can see/use the functions instead of bot permissions setup

internal/discord/handlers.go
OTCompa: Adding a user ID in the `user` field works too!

internal/discord/handlers.go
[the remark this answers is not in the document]
brtran4: ah i see. honestly i would take the route to check to see if they have management/admin role rather than limit visibility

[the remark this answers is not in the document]
brtran4: if you do end up keeping this block, let's add a comment explaining what this block is

internal/discord/handlers.go
OTCompa: Would you prefer multiple fields for minutes/hours/days or just a single field where the input is like "1m", "1h", "1d"?

internal/discord/handlers.go
brtran4: just to be clear is this only a discord id itself or can it also be a nickname?

internal/discord/handlers.go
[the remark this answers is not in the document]
OTCompa: From a brief google search, it seems that I can't limit visibility and usage of the command based on role on this end (at least with discordgo). From what I understand, it can only be done in server settings > integrations > command > role & member overrides, so I believe in our case this would just prevent a random person from using them before the command permissions are set up in the server itself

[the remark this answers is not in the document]
brtran4: ah so this seems to be more of a mod thing @clairefleuret for visibility

internal/discord/handlers.go
clairefleuret: should be  "1m", "1h", "1d"
not multi fields

brtran4: for clarity:
image

internal/discord/handlers.go
brtran4: we should specify in the description that it's discord ID only

internal/discord/command-handlers.go
```diff
@@ -0,0 +1,314 @@
...
+			Description: "Reason for unmute",
+			Required:    true,
+		},
+	},
+}
+
+var BanCommand = &discordgo.ApplicationCommand{
+	Name:                     "ban",
+	DefaultMemberPermissions: &banPermission,
+	Description:              "Ban the specified user.",
```
brtran4, line 123: let's make this similar to `KickCommand` and append ` and notify the user why via DMs.`

brtran4 (changes requested):

internal/discord/command-handlers.go
```diff
@@ -0,0 +1,314 @@
...
+
+// Set up vars for the DefaultMemberPermissions field in each command definition
+var (
+	// I am too dumb and tired to figure out how to set permission restrictions for the roles without these vars.. sorry..
+	kickPermission    int64 = discordgo.PermissionKickMembers
+	mutePermission    int64 = discordgo.PermissionModerateMembers
+	banPermission     int64 = discordgo.PermissionBanMembers
+	channelPermission int64 = discordgo.PermissionManageChannels
+	messagePermission int64 = discordgo.PermissionManageMessages
+)
```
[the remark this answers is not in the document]
OTCompa: It's technically still required because without these, if the bot joins a server with the necessary permissions in the invite link, anyone can use these commands. I've just condensed them into a single admin permission since it's up to the admins how they want to delegate the use of these functions.

internal/discord/command-handlers.go
```diff
@@ -0,0 +1,314 @@
...
+			Type:        discordgo.ApplicationCommandOptionChannel,
+			Name:        "channel",
+			Description: "Channel to purge",
+			Required:    true,
+		},
+		{
+			Type:        discordgo.ApplicationCommandOptionInteger,
+			Name:        "message-number",
+			Description: "Number of messages to purge (100 max)",
+			MaxValue:    100,
```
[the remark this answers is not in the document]
OTCompa: I believe Discord gets mad if we delete a very large number of messages. There's also a [bulk delete function](https://pkg.go.dev/github.com/bwmarrin/discordgo#Session.ChannelMessagesBulkDelete) in discordgo which has a limit of 100, so this number was based off that.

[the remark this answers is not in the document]
brtran4: technically we can raise this limit by sleeping per 100 messages. we could potentially make a new command where it's truly bulk delete but then warn the user that it will take a very long time to purge an entire channel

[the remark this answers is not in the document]
brtran4: as discussed in discord if the # of messages is <= 100, don't worry about sleeping. otherwise set a sleep timer to delete every `n messages % 100` times

internal/discord/command-handlers.go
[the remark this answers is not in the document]
brtran4: ok yeah lets just keep this and delete the comment inside the parenth

internal/discord/command-handlers.go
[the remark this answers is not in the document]
brtran4: closing this so the PR goes in but let's keep track of this :3

brtran4 (approved):

PR #2, opened 2024-05-31
Implement kick function

OTCompa:
Follow up of PR #1, implements the basic functionality for the `/kick` function.
Original PR: https://github.com/OTCompa/moddingway/pull/1

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -51,6 +51,15 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 	}
 }
 
+// CheckUserInGuild checks if the user is in the specified server.
+func (d *Discord) CheckUserInGuild(gid string, user string) error {
+	_, err := d.Session.GuildMember(gid, user)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 var KickCommand = &discordgo.ApplicationCommand{
 	Name:                     "kick",
 	DefaultMemberPermissions: &kickPermission,
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -1,9 +1,13 @@
 package discord
 
 import (
+	"errors"
 	"fmt"
+	"time"
 
 	"github.com/bwmarrin/discordgo"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
 // Set up vars for the DefaultMemberPermissions field in each command definition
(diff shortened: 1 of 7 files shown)


internal/discord/command-handlers.go
```diff
@@ -0,0 +1,396 @@
...
+// CheckUserInGuild checks if the user is in the specified server.
+func (d *Discord) CheckUserInGuild(gid string, user string) error {
+	_, err := d.Session.GuildMember(gid, user)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Discord) CommandLogger(i *discordgo.Interaction) error {
```
brtran4, line 63: same thing here, let's add a doc string. let's also change this func name to `LogCommand`

internal/discord/commands.go
```diff
@@ -0,0 +1,228 @@
...
+		var guildname string
+		guild, err := d.Session.Guild(i.GuildID)
+		if err != nil {
+			fmt.Printf("Unable to find guild name: %v", err)
+			guildname = i.GuildID
+		} else {
+			guildname = guild.Name
+		}
+
+		tempstr := fmt.Sprintf("You are being kicked from %v for the reason:\n%v",
```
brtran4, line 64: can you add screenshots of the kick message here to see what it looks like?

internal/discord/commands.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not kick user <@%v>", userToKick)
+		fmt.Printf("%v: %v\n", tempstr, err)
+
+		err = StartInteraction(s, i.Interaction, tempstr)
+		if err != nil {
+			fmt.Printf("Unable to send ephemeral message: %v\n", err)
+		}
+	} else {
```
brtran4, line 101: i think it makes more sense to move this up in the func. how i picture it is:
1. user uses the command and fills out the fields
2. we dm the user
3. we kick them from the server

brtran4 (changes requested):
some commits from your first PR are present in this PR. is it possible to take them out of this PR?

internal/discord/commands.go
OTCompa: image

OTCompa:
The above commits edits the resulting log message and also sends a followup message whether the user was able to be DMed or not

internal/discord/commands.go
OTCompa: That's the same order unless I'm mistaken.
With the changes below, it should be something along these lines:
1. User uses command
2. Bot logs usage of command 
3. Bot attempts to DM user
4. Bot kicks user regardless 
5. If kicked user was not able to be DMed, inform invoker and edit log message to reflect it

internal/discord/commands.go
brtran4: ok cool this looks good

internal/discord/commands.go
```diff
@@ -1,16 +1,126 @@
...
+
+	// DM the user regarding the kick
+	channel, err := s.UserChannelCreate(userToKick)
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not create a DM with user %v", userToKick)
+		fmt.Printf("%v: %v\n", tempstr, err)
+		dmFailed = true
+	} else {
+		// Get guild name
+		// We could cut down on an API call by hardcoding this
```
brtran4, line 58: let's make a constant file and define the the hardcoded `GUILDNAME` there, then import into this file and other files that we may need in the future

internal/discord/commands.go
brtran4: ah i got error output confused with actual input text. yeah this looks good to me

brtran4 (changes requested):

internal/discord/command-handlers.go
```diff
@@ -32,6 +36,7 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 			PurgeCommand,
 			ExileCommand,
 			UnexileCommand,
+			SetModLogCommand,
```
brtran4, line 39: arguably we should probably move this to its own PR and merge this in first

internal/discord/command-handlers.go
```diff
@@ -46,6 +51,71 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 	}
 }
 
+// CheckUserInGuild checks if the user is in the specified server.
+func (d *Discord) CheckUserInGuild(gid string, user string) error {
```
brtran4, line 55: let's type the full name and do `guild_id`

internal/discord/command-handlers.go
```diff
@@ -46,6 +51,71 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// LogCommand logs the moderation command in the channel specified by LogChannelID
+// It sends an embed with all command arguments as separate fields
+// It additionally returns the sent message in case any edits need to be made
+func (d *Discord) LogCommand(i *discordgo.Interaction) (*discordgo.Message, error) {
```
brtran4, line 66: let's double check that it does not send ephemeral messages

brtran4 (changes requested):

PR #3, opened 2024-06-05
Add warning commands

brtran4:
Added slash commands to warn, clear warnings, delete a warning, and show all warnings for a user, with necessary options and descriptions. Integrated handling of these commands in the InteractionCreate function.

### DIFF
--- a/internal/discord/commands.go
+++ b/internal/discord/commands.go
@@ -110,3 +110,36 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
 func (d *Discord) Unexile(s *discordgo.Session, i *discordgo.InteractionCreate) {
 	return
 }
+
+// Warn attempts to warn a user.
+// fields:
+//
+//	user:		User
+//	reason:		string
+func (d *Discord) Warn(s *discordgo.Session, i *discordgo.InteractionCreate) {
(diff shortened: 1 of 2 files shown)


internal/discord/commands.go
```diff
@@ -110,3 +110,36 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+//
+//	user:		User
+func (d *Discord) ClearWarnings(s *discordgo.Session, i *discordgo.InteractionCreate) {
+	return
+}
+
+// DeleteWarning attempts to delete a warning a user.
+// fields:
+//
+//	warningID:	integer
```
OTCompa, line 134: variable name in the docstring doesn't match the one defined in the slash command definition
Discord also doesn't like uppercase letters in the command argument names so this should be changed instead of the slash command definition
also worth mentioning that I used a hyphen instead of an underscore for one of the argument names in `/purge` `message-number`, I'll fix that if you want to use underscores going forward!

OTCompa (changes requested):

internal/discord/commands.go
brtran4: yea lets go with underscores

OTCompa (approved):

PR #7, opened 2024-06-07
Add function helpers for slash commands

OTCompa:
`mapOptions` creates a map from arguments given in the slash command used.
`ContinueInteraction` sends a follow-up interaction to the user after the initial interaction.
`CheckUserInGuild` checks if the user is in the specified server.


brtran4 (commented):
if you do decide to keep the guild id contained in the other PR, then this looks good to me and i'll approve

brtran4 (approved):

brtran4 (approved):

PR #8, opened 2024-06-07
Implement kick function

OTCompa:
Related to #36 
Dependent on #6 and #7 
To do:
- Resolve current issues
- Update to use logMsg helper functions to fix crashes when mod log channel is not set

### DIFF
--- a/internal/discord/commands.go
+++ b/internal/discord/commands.go
@@ -10,7 +10,94 @@ import (
 //	user: 	User
 //	reason: string
 func (d *Discord) Kick(s *discordgo.Session, i *discordgo.InteractionCreate) {
-	return
+	dmFailed := false
+	optionMap := mapOptions(i)
+
+	// Log usage of command
+	logMsg, err := d.LogCommand(i.Interaction)
+	if err != nil {
+		fmt.Printf("Failed to log: %v\n", err)
+	}
+
+	userToKick := optionMap["user"].UserValue(nil).ID
+
+	// Check if user exists in guild
+	err = d.CheckUserInGuild(i.GuildID, userToKick)
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not kick user <@%v>", userToKick)
+		fmt.Printf("%v: %v\n", tempstr, err)
+
+		err = StartInteraction(s, i.Interaction, tempstr)
+		if err != nil {
+			fmt.Printf("Unable to send ephemeral message: %v\n", err)
+		}
+
+		return
+	}
+
+	// DM the user regarding the kick
+	channel, err := s.UserChannelCreate(userToKick)
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not create a DM with user %v", userToKick)
+		fmt.Printf("%v: %v\n", tempstr, err)
+		dmFailed = true
+	} else {
+		tempstr := fmt.Sprintf("You are being kicked from `%v` for the following reason:\n%v",
+			GuildName,
+			optionMap["reason"].StringValue(),
+		)
+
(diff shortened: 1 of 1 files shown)


internal/discord/commands.go
```diff
@@ -10,7 +10,94 @@ import (
...
+	if len(optionMap["reason"].StringValue()) > 0 {
+		err = d.Session.GuildMemberDeleteWithReason(i.GuildID, userToKick, optionMap["reason"].StringValue())
+	} else {
+		err = StartInteraction(s, i.Interaction, "Please provide a reason for the kick.")
+		if err != nil {
+			fmt.Printf("Unable to send ephemeral message: %v\n", err)
+		}
+
+		return
+	}
```
brtran4, line 68: i think we would rather kick the user first before sending the DM. if the kick fails but the person gets the DM, it would look really confusing to the person

internal/discord/commands.go
```diff
@@ -10,7 +10,94 @@ import (
...
+		err = ContinueInteraction(s, i.Interaction, "Unable to send DM to user.")
+		if err != nil {
+			fmt.Printf("Unable to send ephemeral followup message: %v\n", err)
+		}
+		logMsg.Embeds[0].Description += "\nFailed to notify the user of the kick via DM."
+		_, err = d.Session.ChannelMessageEditEmbed(d.ModLogChannelID, logMsg.ID, logMsg.Embeds[0])
+		if err != nil {
+			fmt.Printf("Unable to edit log message: %v\n", err)
+		}
+	}
```
brtran4, line 100: let's put this right under the DM block

brtran4 (changes requested):

internal/discord/commands.go
```diff
@@ -10,7 +10,94 @@ import (
...
+
+	// DM the user regarding the kick
+	channel, err := s.UserChannelCreate(userToKick)
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not create a DM with user %v", userToKick)
+		fmt.Printf("%v: %v\n", tempstr, err)
+		dmFailed = true
+	} else {
+		tempstr := fmt.Sprintf("You are being kicked from `%v` for the following reason:\n%v",
```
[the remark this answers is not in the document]
OTCompa: image

internal/discord/commands.go
OTCompa: The reason why I didn't do this was cause I'm pretty sure if I attempted to send the DM after the kick, the bot wouldn't share any servers with the kicked user and wouldn't be able to DM them.
I tried anyway and this is what I got:
image

internal/discord/commands.go
OTCompa: This block is here cause I believed it would have been simpler to move a follow-up interaction to the end, rather than adding logic to determine which interaction was first in this function. (`StartInteraction` can only be called once per interaction, any follow-up messages need to use `ContinueInteraction`) I think I could add a helper function that determines which to use if you still want this change though

internal/discord/commands.go
brtran4: ugh thats annoying. ok let's update the error message in the case where we can't kick, the mod should do it manually and _not_ rerun the command. that will send the person two dms which isnt what we want

internal/discord/commands.go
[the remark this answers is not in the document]
brtran4: i think we can go with the quoting to make it clearer

internal/discord/commands.go
brtran4: i think it's worth it to have a helper function. in the future we'll be adding more logic very similarly so it's best to get more of the grunt work out of the way first

PR #9, opened 2024-06-07
Add warning commands

clairefleuret:


brtran4:
you don't need to do this one since it's already covered by https://github.com/naurffxiv/moddingway/pull/3

PR #11, opened 2024-06-08
Add and update function helpers for slash commands

OTCompa:
- `RespondToInteraction` abstracts the need for an if block on whether to use `StartInteraction` or `ContinueInteraction`.
- `parseDuration` parses the duration argument for several slash commands
  - `time.ParseDuration` is insufficient since it does not parse days and more due to design choice
  - 'm' as a unit is parsed as minute (e.g '1m' = 1 minute) instead of month(s) due to the overlap. I favored minutes since it seems to be more common in this situation. Let me know if you want this changed.
- `CheckUserInGuild` is updated to return a `*discordgo.member` object and an error instead of just an error.


internal/discord/commands.go
```diff
@@ -16,6 +19,61 @@ func mapOptions(i *discordgo.InteractionCreate) map[string]*discordgo.Applicatio
...
+	switch groups[2] {
+	case "s":
+		factor = time.Second
+	case "m":
+		factor = time.Minute
+	case "h":
+		factor = time.Hour
+	case "d":
+		factor = time.Hour * 24
+	case "w":
```
brtran4, line 51: is there a reason why we're getting durations based on `seconds`, `days`, `weeks`, and `years` when we only support `minutes`, `hours`, and `days`?

brtran4 (changes requested):

internal/discord/commands.go
OTCompa: Since this was a general function I just implemented what I thought might be used at some point, but I can see that weeks/years would be redundant. Seconds is useful for testing though

brtran4: yeah i think weeks/years would be overkill. let's get rid of those and keep the seconds :)

internal/discord/command-handlers.go
```diff
@@ -126,6 +137,145 @@ func (d *Discord) LogCommand(i *discordgo.Interaction) (*discordgo.Message, erro
...
+	for _, roleToCheck := range rolesToCheck {
+		presentRoles[roleToCheck] = false
+	}
+
+	// check whether user has role
+	for _, role := range member.Roles {
+		if slices.Contains(rolesToCheck, role) {
+			presentRoles[role] = true
+		}
+	}
```
brtran4, line 275: we can condense this into one for loop and check the user's roles against `rolesToCheck`

brtran4 (changes requested):

internal/discord/command-handlers.go
```diff
@@ -126,6 +137,145 @@ func (d *Discord) LogCommand(i *discordgo.Interaction) (*discordgo.Message, erro
...
+}
+
+// UpdateLogMsgTimestamp updates the timestamp for the embed in a logMsg
+func UpdateLogMsgTimestamp(logMsg *discordgo.Message) {
+	if logMsg != nil {
+		logMsg.Embeds[0].Timestamp = time.Now().Format(time.RFC3339)
+	}
+}
```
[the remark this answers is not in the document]
OTCompa: This is mostly for when the logging channel is not set, so it would not have a logMsg to edit

brtran4 (approved):

PR #10, opened 2024-06-08
Get guild roles on initialization

OTCompa:
This PR creates a map of existing roles on bot initialization. It also adds a wait before bot is fully initialized.
Map is structured like so: `Roles[guild_id][role_name]` where the value is a `*discordgo.Role`.

PR #12, opened 2024-06-08
Implement exile and unexile commands

OTCompa:
Dependent on #10 and #11.

### Ephemeral response
image
image

### Mod logging message
image

### User DM
image

### Ephemeral responses
image

### Mod logging message
image


brtran4:
@clairefleuret pls check screenshots. i think we should add more info about what being exiled means in the user DM. basically explaining that they lose access to the server except for the exiled chat etc

internal/discord/commands.go
```diff
@@ -112,7 +112,153 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+	if isExiled {
+		tempstr := fmt.Sprintf("User <@%v> is already exiled", userToExile.ID)
+		RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+		logMsg.Embeds[0].Description += fmt.Sprintf("\n%v", tempstr)
+		_, err = d.Session.ChannelMessageEditEmbed(d.ModLoggingChannelID, logMsg.ID, logMsg.Embeds[0])
+		if err != nil {
+			fmt.Printf("Unable to edit log message: %v\n", err)
+		}
+		return
+	}
```
brtran4, line 159: we can move this chunk to a helper function. return `memberToExile` if they meet the constraints and we find them in the server, else return `nil`

internal/discord/commands.go
```diff
@@ -112,7 +112,153 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+		fmt.Printf("%v: %v\n", tempstr, err)
+		RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+
+		logMsg.Embeds[0].Description += fmt.Sprintf("\n%v", tempstr)
+		_, err = d.Session.ChannelMessageEditEmbed(d.ModLoggingChannelID, logMsg.ID, logMsg.Embeds[0])
+		if err != nil {
+			fmt.Printf("Unable to edit log message: %v\n", err)
+		}
+		return
+	}
```
brtran4, line 189: since we're going to use this logic twice (once here once in `Unexile`, let's make this a helper function for both where we remove one role, and add the other. params could be `roleToRemove` and `roleToAdd` and then add/remove in a general sense

internal/discord/commands.go
```diff
@@ -112,7 +112,153 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+			optionMap["reason"].StringValue(),
+		)
+		_, err = s.ChannelMessageSend(channel.ID, tempstr)
+		if err != nil {
+			tempstr := fmt.Sprintf("Could not send a DM to user <@%v>", userToExile.ID)
+			fmt.Printf("%v: %v\n", tempstr, err)
+			RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+			logMsg.Embeds[0].Description += "\nFailed to notify of the exile via DM"
+		}
+	}
```
brtran4, line 221: we can make a general function for dming someone since we'll be doing this in three commands so far, `Exile`, `Unexile`, and `Kick`. we can pass in a string and have it send the DM

internal/discord/commands.go
```diff
@@ -112,7 +112,153 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+		err = s.GuildMemberRoleAdd(i.GuildID, userToExile.ID, verifiedRole.ID)
+		if err != nil {
+			tempstr := fmt.Sprintf("Could not give user <@%v> role %v", userToExile.ID, verifiedRole.ID)
+			fmt.Printf("%v: %v\n", tempstr, err)
+			logMsg.Embeds[0].Description += tempstr
+		} else {
+			tempstr := fmt.Sprintf("\nUser <@%v> has been successfully unexiled", userToExile.ID)
+			logMsg.Embeds[0].Description += tempstr
+		}
+	}
```
brtran4, line 251: we can call `Unexile` here (or at least split the actual unexile-ing to a helper function, then call it here and the other func)

internal/discord/commands.go
```diff
@@ -121,7 +267,87 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+	if isVerified {
+		tempstr := fmt.Sprintf("User <@%v> is both exiled and verified, nothing has been done", exiledUser.ID)
+		RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+		logMsg.Embeds[0].Description += fmt.Sprintf("\n%v", tempstr)
+		_, err = d.Session.ChannelMessageEditEmbed(d.ModLoggingChannelID, logMsg.ID, logMsg.Embeds[0])
+		if err != nil {
+			fmt.Printf("Unable to edit log message: %v\n", err)
+		}
+		return
+	}
```
brtran4, line 321: how you have it right now is hard to combine with the helper func i suggested in `Exile`. i think we should move this to its own error checking func but it would be even better if we could merge the two into one general error checking func

internal/discord/commands.go
```diff
@@ -121,7 +267,87 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+			tempstr := fmt.Sprintf("Could not give user <@%v> role %v", exiledUser.ID, verifiedRole.ID)
+			fmt.Printf("%v: %v\n", tempstr, err)
+			RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+			logMsg.Embeds[0].Description += fmt.Sprintf("\n%v", tempstr)
+		} else {
+			tempstr := fmt.Sprintf("User <@%v> has been successfully unexiled", exiledUser.ID)
+			RespondToInteraction(s, i.Interaction, tempstr, &isFirst)
+			logMsg.Embeds[0].Description += fmt.Sprintf("\n%v", tempstr)
+		}
+	}
```
brtran4, line 343: same as the `Exile` version of the comment

brtran4 (changes requested):
all of the helper funcs should go in command-handlers

brtran4 (commented):
just checking did you also test minutes, hours, and days as well as testing on multiple users?

clairefleuret:
screenshots for the user are good @brtran4

OTCompa:
> just checking did you also test minutes, hours, and days as well as testing on multiple users?
> 
> let's also test the valid case where not all fields are present. ie: "0m, 0h, 1d" == "1d"

I may have been mistaken, my interpretation of the duration's intended input would just be "1d" and not a combination of multiple units so the code will only consider "1d" as a valid input and not "0m, 0h, 1d". I can easily implement that if needed though!

brtran4:
> > just checking did you also test minutes, hours, and days as well as testing on multiple users?
> 
> > 
> 
> > let's also test the valid case where not all fields are present. ie: "0m, 0h, 1d" == "1d"
> 
> 
> 
> I may have been mistaken, my interpretation of the duration's intended input would just be "1d" and not a combination of multiple units so the code will only consider "1d" as a valid input and not "0m, 0h, 1d". I can easily implement that if needed though!

yeah let's assume that it's both just in case

internal/discord/commands.go
OTCompa: I've been thinking about this for a bit and I decided the way to go about this is to just move this logic (and the similar-ish logic for `/exile`)  into the `ExileUser` and `UnexileUser` helper functions. My reasoning behind these is that the two actions should be paired as an atomic action since I'm always going to check for existing roles before updating the user's roles.

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+// ExileCheckUserHelper checks whether the user meets the requirements to be exiled
+// returns nil if the user does not and returns the member on success
+func (d *Discord) ExileCheckUserHelper(state *InteractionState, userID string) (*discordgo.Member, error) {
+	// Check if user exists in guild
+	member, err := d.GetUserInGuild(state.interaction.GuildID, userID)
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not find user <@%v> in guild", userID)
+		fmt.Printf("%v: %v\n", tempstr, err)
+		RespondToInteraction(state.session, state.interaction.Interaction, tempstr, &state.isFirst)
+		return nil, err
```
brtran4, line 492: let's check they already don't have the exile role

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+	}
+	return member, nil
+}
+
+// ExileRoleHelper removes a role `roleIDToRemove` and then adds a role `roleIDToAdd` to the user `userID`
+func (d *Discord) ExileRoleHelper(state *InteractionState, userID string, roleIDToRemove string, roleIDToAdd string) error {
```
brtran4, line 498: let's rename this to be a bit more specific like `RoleRemoveAddHelper`. this isn't specific to exile so we can make this general while also being more descriptive about the function

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+	// Attempt to remove role first
+	err := state.session.GuildMemberRoleRemove(state.interaction.GuildID, userID, roleIDToRemove)
```
brtran4, line 500: let's check to make sure they have the role first

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+	// Abort entire process if role removal fails
+	if err != nil {
+		tempstr := fmt.Sprintf("Could not remove role %v from user <@%v>", roleIDToAdd, userID)
+		fmt.Printf("%v: %v\n", tempstr, err)
+		RespondToInteraction(state.session, state.interaction.Interaction, tempstr, &state.isFirst)
+		AppendLogMsgDescription(state.logMsg, tempstr)
+		return err
+	} else {
+		// Add role
```
brtran4, line 509: likewise, check to see if they already have the role. we should probably move the checks to a helper and call that helper at the beginning of the func

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+	exileRole := d.Roles[state.interaction.GuildID]["Exiled"]
+	verifiedRole := d.Roles[state.interaction.GuildID]["Verified"]
+	presentRoles := CheckUserForRoles(member, []string{exileRole.ID})
+
+	if presentRoles[exileRole.ID] {
+		tempstr := fmt.Sprintf("User <@%v> is already exiled", member.User.ID)
+		RespondToInteraction(state.session, state.interaction.Interaction, tempstr, &state.isFirst)
+		AppendLogMsgDescription(state.logMsg, tempstr)
+		d.EditLogMsg(state.logMsg)
+		err := fmt.Errorf("already exiled")
```
brtran4, line 533: we could move this error checking to the helper from above

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+		} else {
+			return nil
+		}
+	}
+}
+
+func (d *Discord) ExileUser(state *InteractionState, member *discordgo.Member, reason string) error {
+	exileRole := d.Roles[state.interaction.GuildID]["Exiled"]
+	verifiedRole := d.Roles[state.interaction.GuildID]["Verified"]
+	presentRoles := CheckUserForRoles(member, []string{exileRole.ID})
```
brtran4, line 526: we can make a getter for these three and use them here and the other func

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,96 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+	}
+
+	if presentRoles[verifiedRole.ID] {
+		tempstr := fmt.Sprintf("User <@%v> is both exiled and verified, nothing has been done", member.User.ID)
+		RespondToInteraction(state.session, state.interaction.Interaction, tempstr, &state.isFirst)
+		AppendLogMsgDescription(state.logMsg, tempstr)
+		d.EditLogMsg(state.logMsg)
+		err := fmt.Errorf("both exiled and verified")
+		return err
+	}
```
brtran4, line 562: same as the other comment, we can make this into one general function

internal/discord/commands.go
```diff
@@ -112,7 +112,98 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+		return
+	}
+
+	userToExile := optionMap["user"].UserValue(nil)
+
+	// Check if user meets the requirements for an exile
+	member, err := d.ExileCheckUserHelper(state, userToExile.ID)
+	if err != nil {
+		return
+	}
```
brtran4, line 142: ehhhh if we're already error checking in the helper funcs we might as well get rid of it here

internal/discord/commands.go
```diff
@@ -112,7 +112,98 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+			d.ModLoggingChannelID,
+			logMsg.Embeds[0],
+		)
+		if err != nil {
+			fmt.Printf("Failed to log: %v\n", err)
+		}
+	}
+
+	// Check if user meets the requirements
+	member, err = d.ExileCheckUserHelper(state, userToExile.ID)
```
brtran4, line 188: same as above, we don't need to check 3-4 times if they are eligible for exile

internal/discord/commands.go
```diff
@@ -121,7 +212,40 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+		isFirst:     true,
+	}
+
+	exiledUser := optionMap["user"].UserValue(nil)
+
+	// Check if user meets the requirements for unexile
+	member, err := d.ExileCheckUserHelper(state, exiledUser.ID)
+	if err != nil {
+		return
+	}
```
brtran4, line 234: same as earlier comment

brtran4 (changes requested):
let's try to lower the amount of code duplication and unnecessary error checking in this PR

brtran4:
now that im looking at this more im kinda contradicting a little bit lol but basically we should either go with

1. keep the error checking in the helper funcs so we dont have to do it in the main funcs
or
2. keep error checking only in the main funcs

both seem good to me but we are kinda dipping into both at the same time, and we should keep it explicitly to only one option

internal/discord/command-handlers.go
OTCompa: Not checking for the roles in this block of code is intentional (related to this comment https://github.com/naurffxiv/moddingway/pull/12#discussion_r1632511083) it is done later down the line

clairefleuret:
Adding a note when a duration IS NOT given, the user is dm'd they are INDEFINTIELY exiled.

Additionally, link this form: https://dyno.gg/form/86459368

internal/discord/command-handlers.go
```diff
@@ -479,3 +479,70 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+		fmt.Printf("Unable to unexile user <@%v>: %v\n", userID, err)
+		return err
+	}
+
+	// Remove verified role and add exiled role
+	return d.RoleRemoveAddHelper(state, userID, roleToRemove, roleToAdd)
+}
+
+func (d *Discord) UnexileUser(state *InteractionState, userID string, reason string) error {
+
```
brtran4, line 526: nit: remove extra line

internal/discord/commands.go
```diff
@@ -112,7 +112,73 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+	// DM user regarding the exile, doesn't matter if DM fails
+	tempstr = fmt.Sprintf("You are being exiled from `%v` until <t:%v> for the following reason:\n> %v",
+		GuildName,
+		endTime.Unix(),
+		optionMap["reason"].StringValue(),
+	)
+	d.SendDMToUser(state, userToExile.ID, tempstr)
+	d.EditLogMsg(logMsg)
+
+	time.Sleep(duration)
```
brtran4, line 157: since the reqs changed, let's not even bother calculating the time if theres no duration arg to do so. that said we would probably need to have another tempstr case where there is no duration arg

internal/discord/commands.go
```diff
@@ -112,7 +112,73 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+
+	// Reuse the original embed format but clear existing info
+	if logMsg != nil {
+		logMsg.Embeds[0].Description = ""
+	}
```
brtran4, line 162: i can see us using this a lot in the future. maybe move this to a helper function?

brtran4 (changes requested):
oof yeah i can definitely see `commands.go` being bloated after we implement all of these commands (https://github.com/naurffxiv/moddingway/issues/16). let's keep this in mind so that when we refactor it isn't as much of a pain as it oculd be

internal/discord/command-handlers.go
```diff
@@ -713,3 +729,69 @@ func (d *Discord) InteractionCreate(s *discordgo.Session, i *discordgo.Interacti
...
+		} else {
+			return nil
+		}
+	}
+}
+
+func (d *Discord) ExileUser(state *InteractionState, userID string, reason string) error {
+	// Check user for specified roles
+	roleToRemove := "Verified"
+	roleToAdd := "Exiled"
```
brtran4, line 763: ehhhh we should probably make an issue since we want this in PR to go in soon, but we should probably move this to the constants file

internal/discord/commands.go
```diff
@@ -103,7 +103,93 @@ func (d *Discord) Purge(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+			return
+		}
+	}
+
+	userToExile := optionMap["user"].UserValue(nil)
+
+	err = d.ExileUser(state, userToExile.ID, optionMap["reason"].StringValue())
+	if err != nil {
+		return
+	}
```
brtran4, line 134: it's unsettling have this in between two `if hasDuration` blocks.... do we have to calculate the duration before calling `ExileUser`? i think we would rather have it be together in the if block below

brtran4 (changes requested):

internal/discord/commands.go
OTCompa: I thought about that, but the issue is if the invoker formatted the duration argument incorrectly it would exile the user first then fail before carrying anything else out.

brtran4: I feel like there is a better way around this 🤔 lets open a new tech debt issue to deal with this and get this code merged in

brtran4 (approved):

PR #13, opened 2024-06-09
Disable non-implemented commands for launch

brtran4:
As we are getting ready to launch the bot, we need to only show commands that are currently functional. This commit disables all commands that we are not planning on releasing on initial launch.

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -1,14 +1,24 @@
 package discord
 
 import (
-	"errors"
 	"fmt"
+	"regexp"
+	"slices"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/bwmarrin/discordgo"
 	"golang.org/x/text/cases"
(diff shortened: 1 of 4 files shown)


internal/discord/command-handlers.go
```diff
@@ -24,23 +24,23 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
...
+			// SlowmodeOffCommand,
+			// PurgeCommand,
 			ExileCommand,
 			UnexileCommand,
-			SetModLoggingCommand,
-			AddWarningCommand,
-			ClearWarningsCommand,
-			DeleteWarningCommand,
-			ShowAllWarningsCommand,
+			// SetModLoggingCommand,
```
OTCompa, line 39: SetModLoggingCommand is implemented

brtran4: yeah sure we can add it in

OTCompa (approved):

PR #14, opened 2024-06-09
Update command options for Exile and Unexile commands

brtran4:
This commit changes the exile duration to optional.

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -566,15 +566,15 @@ var ExileCommand = &discordgo.ApplicationCommand{
 		},
 		{
 			Type:        discordgo.ApplicationCommandOptionString,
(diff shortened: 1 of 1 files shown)


clairefleuret (commented):
@brtran4 have we implemented a Direct Message system yet or are these just the commands

brtran4:
> @brtran4 have we implemented a Direct Message system yet or are these just the commands

that's in https://github.com/naurffxiv/moddingway/pull/12

clairefleuret:
> > @brtran4 have we implemented a Direct Message system yet or are these just the commands
> 
> that's in #12

lets make sure when a duration IS NOT selected, the user is notified they are inefficiently exiled. other than that that this is good.

ill make a comment in the other pr

clairefleuret (approved):

PR #17, opened 2024-06-22
NFC: Remove unused time import from commands.go

brtran4:


OTCompa:
This will be used in #12. I probably included it in the wrong PR, oops. I can move this line of code there if needed

brtran4:
> This will be used in #12. I probably included it in the wrong PR, oops. I can move this line of code there if needed

Yeah we should. If we run a linter on the current code, it will fail, so we will need to do that

brtran4:
i guess since the code is in there shouldnt be any linting problems. closing

PR #25, opened 2024-07-03
Create Postgres in docker and initialize tables

KingPieson:
Create local postgres instance via Docker for local development. Create users and strikes tables in DB init file

### DIFF
--- a/postgres/create_tables.sql
+++ b/postgres/create_tables.sql
@@ -1,0 +1,22 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS users (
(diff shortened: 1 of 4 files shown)


postgres/create_tables.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS users (
+	userID INT GENERATED ALWAYS AS IDENTITY,
```
brtran4, line 4: how does this exactly index the users? are we taking the discord-indexed ids or will the users just index themselves?

brtran4 (approved):

PR #26, opened 2024-07-13
Allow SetModLoggingChannel to send embed logs.

### LINKED ISSUE #24: When setting a mod log channel, change the confirmation to a message the channel set
Would be a better implementation if we have a confirmation message in the channel rather than an echo.

brtran4:
Closes #24

### DIFF
--- a/internal/discord/commands.go
+++ b/internal/discord/commands.go
@@ -234,7 +234,17 @@ func (d *Discord) SetModLoggingChannel(s *discordgo.Session, i *discordgo.Intera
 	channelID := options[0].ChannelValue(nil).ID
 	d.ModLoggingChannelID = channelID
 
+	logMsg, _ := d.LogCommand(i.Interaction)
+	state := &InteractionState{
+		session:     s,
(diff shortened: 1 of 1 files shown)


brtran4:
the new logging embed will look like the below

Screenshot 2024-07-12 at 9 10 12 PM

OTCompa:
I think what Claire wants in the issue is that if `/setmodlog` is used in a different channel than the channel specified in the argument (e.g `/setmodlog #logs` is used in #general), it should send a confirmation message in the specified channel and not the current channel. We've had 2 separate instances where was 1. lacking permission to post and 2. set to a category instead of the channel so I'm assuming that's probably what she wants. I believe this change sends a log message in the channel the command was invoked in which isn't the same?

brtran4:
i shouldn't have cut off the top part, but yeah that's exactly what it's doing. the command was used in `#general` but the log appeared in `#test`
Screenshot 2024-07-17 at 2 00 31 PM

OTCompa (approved):

PR #27, opened 2024-07-13
Add dockerfile for moddingway app

OTCompa:
`docker-compose.yml` will check a local file named `.env` for the `DISCORD_TOKEN` (and other secrets if needed in the future)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,8 +7,8 @@ services:
     ports:
       - 5432:5432
     environment:
-      - POSTGRES_PASSWORD=S3cret
-      - POSTGRES_USER=test_user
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -3,10 +3,17 @@ services:
     container_name: postgres_db
     build:
       context: .
-      dockerfile: ./docker/postgres.dockerfile
+      dockerfile: ./docker/Postgres.Dockerfile
     ports:
       - 5432:5432
     environment:
       - POSTGRES_PASSWORD=S3cret
```
brtran4, line 10: while we're here we might as well not expose the password

brtran4 (approved):
approving on behalf of @KingPieson

PR #29, opened 2024-07-13
Limit creation of commands to a single specified guild

### LINKED ISSUE #21: Limit creation of commands to only NAUR specifically
Since this bot is meant to be internal, commands should only be created in NAUR. It also presents a possible security issue if the bot's permissions on the account side is not configured properly.
For example, `ModLoggingChannelID` is "shared" across all guilds since it's only meant to be used in 1 guild. If the account that the bot runs on is configured to be a public bot where anyone can invite it, they could theoretically route the log messages to another server. 
The current "live" bot is appropriately set to private so nobody can invite it but I believe it's better to address this in code.

OTCompa:
Dependent on #28 
Resolves #21 
Default specified guild is set to NAUR
Leaving this as draft until #28 is merged to prevent messy commits

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -28,11 +28,12 @@ var (
 // AddCommands registers the slash commands with Discord
 func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 	fmt.Printf("Initializing Discord...\n")
+	foundTargetGuild := false
 	for _, discordGuild := range event.Guilds {
 		if discordGuild.ID != d.GuildID {
 			continue
 		}
-
+		foundTargetGuild = true
 		// Adding commands to a list to prepare in bulk
 		var commands []*discordgo.ApplicationCommand
 		commands = append(commands,
@@ -64,6 +65,11 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 		if err != nil {
 			fmt.Printf("Could not add some commands: %v \n", err)
 		}
+	}
+
+	if !foundTargetGuild {
+		d.Session.Close()
+		panic("Bot is not present in the specified guild.")
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/discord/command-handlers.go
```diff
@@ -30,37 +30,39 @@ func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
...
-
-		fmt.Printf("Adding commands...\n")
-		commandList, err := s.ApplicationCommandBulkOverwrite(event.User.ID, discordGuild.ID, commands)
-		fmt.Printf("List of successfully created commands:\n")
-		for _, command := range commandList {
-			fmt.Printf("\t%v\n", command.Name)
-		}
-		if err != nil {
-			fmt.Printf("Could not add some commands: %v \n", err)
+		if discordGuild.ID == d.GuildID {
```
brtran4, line 33: for better min-diff and clearer error checking let's do the inverse

```
if discordGuild.ID == d.GuildID {
   raise error
}
```

brtran4 (changes requested):

internal/discord/command-handlers.go
```diff
@@ -28,8 +28,10 @@ var (
 // AddCommands registers the slash commands with Discord
 func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 	fmt.Printf("Initializing Discord...\n")
-
 	for _, discordGuild := range event.Guilds {
+		if discordGuild.ID != d.GuildID {
+			continue
```
brtran4, line 33: ah i missed the forloop above. hmm i'm thinking we would still want to error if we can't find ANY matching guild ids

internal/discord/command-handlers.go
```diff
@@ -28,9 +28,12 @@ var (
...
 func (d *Discord) AddCommands(s *discordgo.Session, event *discordgo.Ready) {
 	fmt.Printf("Initializing Discord...\n")
-
+	foundTargetGuild := false
 	for _, discordGuild := range event.Guilds {
-
+		if discordGuild.ID != d.GuildID {
+			continue
+		}
+		foundTargetGuild = true
```
brtran4, line 36: this should be in an else block

brtran4 (changes requested):

internal/discord/command-handlers.go
brtran4: ah i meant just that line, but this approach works too. we'll need to revert the inverse and just do `if discordGuild.ID == d.GuildID`

brtran4 (approved):

PR #30, opened 2024-07-13
Make testing the bot easier

OTCompa:
Part of #50 
Dependent on #27 and #29
Adds a way to easily override the default guild ID and log channel ID set in `const.go` through environment variables to test the bot on a personal server without having to change it every time.
To use, set `DEBUG` to true in `.env` and the corresponding `GUILD_ID` and `MOD_LOGGING_CHANNEL_ID`
Leaving this as draft until #27 and #29 is merged to prevent messy commits (will be redone when they are merged)


brtran4:
we should create a README that has this information from the commit message

brtran4 (approved):

PR #32, opened 2024-07-20
Implement ban command

brtran4:


internal/discord/commands.go
```diff
@@ -41,7 +41,55 @@ func (d *Discord) Unmute(s *discordgo.Session, i *discordgo.InteractionCreate) {
 //	user:		User
 //	reason:		string
 func (d *Discord) Ban(s *discordgo.Session, i *discordgo.InteractionCreate) {
-	return
+	optionMap := mapOptions(i)
+	logMsg, _ := d.LogCommand(i.Interaction)
+
+	state := &InteractionState{
+		session: 	 s,
+		interaction: i,
```
brtran4, line 49: this is definitely a problem since it's properly formatted on my IDE, but not here

internal/discord/commands.go
```diff
@@ -41,7 +41,55 @@ func (d *Discord) Unmute(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+
+	tempstr := fmt.Sprintf("<@%v> has been banned", userToBan)
+	RespondAndAppendLog(state, tempstr)
+	d.EditLogMsg(logMsg)
+
+	// Attempt to ban user
+	if len(optionMap["reason"].StringValue()) > 0 {
+		err = d.Session.GuildBanCreateWithReason(i.GuildID, userToBan, optionMap["reason"].StringValue(), 0)
+	} else {
+		err = StartInteraction(s, i.Interaction, "Please provide a reason for the ban.")
```
OTCompa, line 86: We should try to use `RespondToInteraction()` instead. Discord differentiates between the first respond and any further follow-ups, so any use of `StartInteraction()` after any response will not do anything.

OTCompa (changes requested):

OTCompa (approved):

PR #53, opened 2024-07-23
move exiled and verified role strings to constants

KingPieson:


KingPieson:
I'm unsure what "buildx failed with: ERROR: denied: installation not allowed to Write organization package" in the job means here, possibly I configured the PR wrong?

brtran4 (approved):

brtran4:
noted in the discord

> I don't think this is fixable actually. The fork is outside the organization so it can't upload a package to the organization
It should work on a branch inside the same fork. So clone `naurffxiv/moddingway` and create branches there and submit PRs

PR #54, opened 2024-07-23
Make getting environment variables cleaner

OTCompa:
Currently for every environment variable we get, we need to error check like so:
```golang
discordToken, ok := os.LookupEnv("DISCORD_TOKEN")
if !ok {
  panic("You must supply a DISCORD_TOKEN to start!")
}

guildID, ok := os.LookupEnv("GUILD_ID")
if !ok {
  panic("You must supply a GUILD_ID to start!")
}

modLoggingChannelID, ok := os.LookupEnv("MOD_LOGGING_CHANNEL_ID")
if !ok {
  panic("You must supply a MOD_LOGGING_CHANNEL_ID to start!")
}

// rest of the env vars
```

This helper will allow us to simplify it slightly and move error checking to right before we start any connections.
```golang
discordToken := eg.GetEnv("DISCORD_TOKEN")
guildID := eg.GetEnv("GUILD_ID")
modLoggingChannelID := eg.GetEnv("MOD_LOGGING_CHANNEL_ID")

// rest of the env vars

if !eg.Ok {
  tempstr := fmt.Sprintf("You must supply a %s to start!", eg.EnvName)
  panic(tempstr)
}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,7 +19,6 @@ func main() {
 	d := &discord.Discord{}
 
 	discordToken := env.GetEnv("DISCORD_TOKEN")
-	discordToken = strings.TrimSpace(discordToken)
 
 	debug := env.GetEnv("DEBUG")
 	debug = strings.ToLower(debug)
--- a/main.go
+++ b/main.go
@@ -8,36 +8,35 @@ import (
 	"syscall"
 
 	"github.com/naurffxiv/moddingway/internal/discord"
+	"github.com/naurffxiv/moddingway/internal/util"
 )
 
 func main() {
-	discordToken, ok := os.LookupEnv("DISCORD_TOKEN")
-	if !ok {
-		panic("You must supply a DISCORD_TOKEN to start!")
+	env := util.EnvGetter{
+		Ok: true,
 	}
-	discordToken = strings.TrimSpace(discordToken)
 
 	d := &discord.Discord{}
 
-	debug, _ := os.LookupEnv("DEBUG")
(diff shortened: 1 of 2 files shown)


brtran4:
will this PR include actually using that? or will that be in a chained PR?

OTCompa:
> will this PR include actually using that? or will that be in a chained PR?

It'll be in a chained PR. I didn't want to make the PR further down the line too big so I figured I'd split it into smaller one

brtran4:
> > will this PR include actually using that? or will that be in a chained PR?
> 
> It'll be in a chained PR. I didn't want to make the PR further down the line too big so I figured I'd split it into smaller one. I can also update the current `main.go` in this PR too since it'll have to be done later anyway if that's cool

i think the change is small enough where we can fit it into one PR

main.go
```diff
@@ -8,30 +8,25 @@ import (
...
+	env := util.EnvGetter{
+		Ok: true,
 	}
-	discordToken = strings.TrimSpace(discordToken)
 
 	d := &discord.Discord{}
 
-	debug, _ := os.LookupEnv("DEBUG")
+	discordToken := env.GetEnv("DISCORD_TOKEN")
+	discordToken = strings.TrimSpace(discordToken)
```
brtran4, line 22: we could just move this into `GetEnv` since all strings will hit this case

brtran4 (changes requested):

brtran4 (approved):

PR #55, opened 2024-07-23
Add exiles table to database schema

OTCompa:
This PR adds the `exiles` table to the database schema. It will be used to check when people should be automatically unexiled after the exile duration is over. It also fixes some minor formatting/spelling issues in the original SQL files.

The `exileStatus` column will be defined using an enum that's included in this PR. The names for each enum aren't really that good, so feel free to suggest better ones.
- `IndefiniteExile` means a duration was not set for the exile
- `TimedExile` means a current exile that either is still in effect or has yet to be unexiled
- `Unknown` means there was some error encountered along the way (probably when performing the unexile)


brtran4:
> This PR adds the `exiles` table to the database schema. It will be used to check when people should be automatically unexiled after the exile duration is over. It also fixes some minor formatting/spelling issues in the original SQL files.
> 
> The `exileStatus` column will be defined using an enum that's included in this PR. The names for each enum aren't really that good, so feel free to suggest better ones.
> 
> * `Unexiled` means the user has been unexiled already.
> * `IndefiniteExile` means a duration was not set for the exile
> * `TimedExile` means a current exile that either is still in effect or has yet to be unexiled
> * `TimedExileMissing` means a user was exiled with a duration, but they have left the server before they were unexiled (to prevent repeated attempts to unexile them when they aren't in the server)

i think we should drop the entry if the user has been unexiled. @clairefleuret do we want to store the amount of exiles a person accumulated?

just a funny thought, but we can rename `TimedExileMissing` to `Runaway` since they're "running" away from the exile 😝

clairefleuret:
> i think we should drop the entry if the user has been unexiled. @clairefleuret do we want to store the amount of exiles a person accumulated?

Nah, strikes being tracked is what we care about.

brtran4 (changes requested):
yeah let's drop the entries if they are unexiled

brtran4 (approved):

PR #56, opened 2024-07-23
Add Unban functionality

### LINKED ISSUE #37: Basic Moderation - Ban Command
Ban - Ability to ban/unban. Notify user reason via dms
	/ban <user> <reason>
	/unban <user> <reason>

brtran4:
Closes #37

### DIFF
--- a/internal/discord/commands.go
+++ b/internal/discord/commands.go
@@ -63,7 +63,7 @@ func (d *Discord) Ban(s *discordgo.Session, i *discordgo.InteractionCreate) {
 		if err != nil {
 			fmt.Printf("Unable to send ephemeral message: %v\n", err)
 		}
-		
+
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


internal/discord/commands.go
```diff
@@ -98,7 +98,53 @@ func (d *Discord) Ban(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+		return
+	}
+
+	// Dm the user regarding the unban
+	// unbanstr := fmt.Sprintf("You are being unbanned from `%v` for the following reason:\n> %v", GuildName)
+	// d.SendDMToUser(state, userToUnban, unbanstr)
+	//
+	// tempstr := fmt.Sprintf("<@%v> has been unbanned", userToUnban)
+	// RespondAndAppendLog(state, tempstr)
+	// d.EditLogMsg(logMsg)
```
brtran4, line 134: @clairefleuret the bot can't message the person because they don't share the same server. is DMing the user a requirement?

internal/discord/commands.go
```diff
@@ -98,7 +98,53 @@ func (d *Discord) Ban(s *discordgo.Session, i *discordgo.InteractionCreate) {
 //	user:		User
 //	reason:		string
 func (d *Discord) Unban(s *discordgo.Session, i *discordgo.InteractionCreate) {
-	return
+	optionMap := mapOptions(i)
```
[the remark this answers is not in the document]
OTCompa: I think you can use [this](https://pkg.go.dev/github.com/bwmarrin/discordgo#Session.GuildBans) to access the ban list then use the argument's autocomplete field to suggest the ban list. Not entirely sure if this will work, but could be worth a try

internal/discord/commands.go
clairefleuret: In that case, no. We will have to do it manually. If anyone is requesting an unban, they would need to be in contact with us anyways.

PR #59, opened 2024-07-23
Add database module

OTCompa:
This module will house any methods/functions that interact with the database to keep it separate. Any queries required by a feature will be added by their respective PRs in this module, so it only contains a connect function right now. (I'm not entirely sure if this is the "proper" way of doing things, so let me know if there's a better way to organize things)


internal/database/exiles.go
```diff
@@ -0,0 +1,132 @@
...
+	ORDER BY exileID DESC
+	LIMIT 1`
+
+	var exileID int
+	err := conn.QueryRow(context.Background(), query, dbUserID).Scan(&exileID)
+	if err != nil {
+		return -1, err
+	}
+	return exileID, nil
+}
```
brtran4, line 115: is there a case where we need this?

internal/database/users.go
```diff
@@ -0,0 +1,43 @@
...
+	WHERE discordUserId = $1 AND discordGuildId = $2
+	LIMIT 1`
+
+	var dbUserID int
+	err := conn.QueryRow(context.Background(), query, discordUserID, discordGuildID).Scan(&dbUserID)
+	if err != nil {
+		return -1, err
+	}
+	return dbUserID, nil
+}
```
brtran4, line 31: i think it would be better (coding principle wise) to split these into two. we shouldn't combine a getter and setter together

brtran4 (changes requested):

internal/database/exiles.go
OTCompa: It's used someone calls `/unexile` to remove the currently active exile
I could've used the first exile found, but I went with most recent in case someone gets exiled again after a failed exile where the exileStatus is set to `Unknown`.

brtran4: if it's unknown shouldn't we just drop the entry? otherwise that sounds good to me

OTCompa: I think I'd want to keep it in for debugging purposes (and to figure out if anyone was somehow left dangling)

brtran4: if we want to use it for debugging purposes, we should move this to a test file (like `exiles-tests.go`) and make a unit test for it. if/when we need to debug this, we can just call this in the shell instead of actually incorporating our prod code with it

OTCompa: There's a small misunderstanding, I meant keeping the DB entry for when something wrong happens as opposed to dropping the entry.

brtran4: im still kinda against that because we shouldnt be putting the database in a state where we have to use it to debug. that responsibility is on the code accessing the database, and not the database itself

OTCompa: Alright, I guess I can just look at stdout if there are any issues, I'll go ahead and change this to GetUserExile()

brtran4 (approved):

Issue #51, opened 2024-07-21
Separate worker to handle timed exiles


OTCompa:
Also remove timed exile logic from the `/exile` command

Issue #19, opened 2024-06-28
Re-organize code in Exile command


OTCompa:
Currently, there are 2 `if hasDuration` blocks broken up code to exile the user, which there is probably a better way around.
We could just combine both if blocks and move exiling the user to within the blocks to make the code more readable at the expense of some duplicated code.

clairefleuret:
Adding task to re-organize how logging works.

Instead of being dynamically set, we can have it be static.

Channel ID: 1172324840947056681

brtran4:
> Adding task to re-organize how logging works.
> 
> Instead of being dynamically set, we can have it be static.
> 
> Channel ID: 1172324840947056681

@clairefleuret we should split this to be its own issue for logging. this is specifically for the exile/unexile code

PR #70, opened 2024-07-24
Add separate worker to handle timed exiles

OTCompa:
Dependent on #59 and part of #51. Paired with #71.
This PR adds a separate worker that checks the database for currently exiled users whose exile duration has expired.

`temp.go` holds some functions that will replace the functions with a similar name when I begin refactoring. I've included them in this PR since they're needed by the worker


internal/discord/temp.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/bwmarrin/discordgo"
+)
+
+// Functions/methods that are meant to replace the existing non-temp versions during refactoring
```
brtran4, line 11: let's attach the issue number ([#19](https://github.com/naurffxiv/moddingway/issues/19))

brtran4 (approved):

PR #71, opened 2024-07-24
Connect main app to the database

OTCompa:
Dependent on #59 and part of #51. Paired with #70.
 This PR connects the main app to the database and adds the corresponding entries when exiling a user.


internal/discord/commands.go
```diff
@@ -245,7 +252,34 @@ func (d *Discord) Exile(s *discordgo.Session, i *discordgo.InteractionCreate) {
...
+				fmt.Printf("%v: %v\n", tempstr, err)
+				RespondAndAppendLog(state, tempstr)
+				return
+			}
+		}
+
+		exileEntryArgs := database.AddExileEntryArgs{
+			DbUserID: dbUserID,
+			Reason: optionMap["reason"].StringValue(),
+			ExileStatus: enum.TimedExile,
```
brtran4, line 273: Is this supposed to be UntimedExile?

main.go
```diff
@@ -34,10 +35,19 @@ func main() {
 		d.Init(discordToken)
 	}
 
+	dbArgs := database.DbInfo{
+		Host: env.GetEnv("POSTGRES_HOST"),
```
brtran4, line 39: Let's update the README to include these new postgres env vars

brtran4 (changes requested):

main.go
OTCompa: This shouldn't need to be defined in the .env file since the database we're using is also running in the same docker container. These are hardcoded in `docker-compose.yml`

brtran4: Ok that sounds good to me

brtran4 (approved):

PR #86, opened 2024-08-21
Removed/Commented Out the unneeded Commands. Fixes #84

### LINKED ISSUE #84: Remove unneeded commands
we should remove the following commands and code stubs

- `/slowmode`
- `/slowmodeoff`
- `/kick`

MachoMaheen:
Fixed the Issue #84 . This PR removes the following commands and code stubs:
 - /slowmode
- /slowmodeoff
- /kick

### DIFF
--- a/internal/discord/command-handlers.go
+++ b/internal/discord/command-handlers.go
@@ -384,25 +384,25 @@ func ClearEmbedDescription(logMsg *discordgo.Message) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


brtran4 (changes requested):
in addition to removing the command entirely, we should also remove the commands in `AddCommands`

brtran4:
don't worry about the docker job, it won't run on your fork. the more important part is to make sure that the golangci-lint job passes

PR #90, opened 2024-08-24
Add make clean to Makefile

brtran4:
As we keep running these build commands, many docker images start dangling without an attached container. Eventually this takes too much space.

This commit adds an option to "clean" out the docker images with `make clean`.


brtran4:
my git is being really weird so im going to close this and re open

Issue #72, opened 2024-07-24
Basic Moderation - Recruitment Auto Mod


clairefleuret:
We need a module that will auto delete posts after a certain amount of time in the recruitment channels. Another condition for deleting is if the original message was deleted.

Upon deletion, a log should be sent to mod actions channel 1172324840947056681. Additionally, have Moddingway DM the user their post has been removed for the reason.

Looking for Group Channel ID: 1273261496968810598
Delete after 30 days of inactivity

Looking for More Channel ID: 1273261496968810598 
Delete after 30 days of inactivity

Temporary Recruitment Channel ID: 1240356145311252615
Delete after 15 days of inactivity

![Image](https://github.com/user-attachments/assets/d1c160c7-17ae-442a-9d58-68aeb68a64fe)

From some people I asked how they would do this:

> An easy way would have to create an entry in a database everytime a forum is created and update the timestamp everytime someone sends a message in it. Get a daily task to go through the timestamps in the database and delete all forums entries that are 30+ days.

> Discord.py: Every X time, loop through all the [threads](https://discordpy.readthedocs.io/en/latest/api.html#discord.ForumChannel.threads) in the channel and get the [thread history](https://discordpy.readthedocs.io/en/latest/api.html#discord.Thread.history). Compare the most recent message’s [created at](https://discordpy.readthedocs.io/en/latest/api.html#discord.Message.created_at) date and time and if it is older than 30 days, you can [delete](https://discordpy.readthedocs.io/en/latest/api.html#discord.Thread.delete) the thread. No database needed. I’m sure Discord.go has the same functions, although I’m not sure why you wouldn’t want to just lock and archive them.

PR #96, opened 2024-08-27
Add DMing user back to auto-unexile

### LINKED ISSUE #95: Autounexile should DM the user

OTCompa:
Resolves #95 
Since I was here, I also deferred removing the DB entry to ensure the exile entry will get removed from the database even on errors. This will ensure that we don't try to unexile again the next time around if the exile somehow fails.

### DIFF
--- a/internal/discord/temp.go
+++ b/internal/discord/temp.go
@@ -100,3 +100,14 @@ func CreateMemberEmbed(member *discordgo.Member, description string, footer stri
 		}
 	}
 }
+
+func (d *Discord) TempSendDMToUser(userID string, message string) error {
+	// Open DM channel with user
+	channel, err := d.Session.UserChannelCreate(userID)
+	if err != nil {
+		return err
+	} else {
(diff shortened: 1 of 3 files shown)


internal/discord/temp.go
```diff
@@ -100,3 +100,14 @@ func CreateMemberEmbed(member *discordgo.Member, description string, footer stri
...
+
+func (d *Discord) TempSendDMToUser(userID string, message string) error {
+	// Open DM channel with user
+	channel, err := d.Session.UserChannelCreate(userID)
+	if err != nil {
+		return err
+	} else {
+		_, err = d.Session.ChannelMessageSend(channel.ID, message)
+		return err
+	}
```
brtran4, line 112: let's move this to an entirely new file called `util.go` i will be needing this for https://github.com/naurffxiv/moddingway/issues/72

brtran4 (changes requested):

brtran4 (approved):

PR #103, opened 2024-09-03
add methods to interact with strikes db

KingPieson:
Add strikes DB table, create methods to interact with table, add strikes command (still hidden)

### DIFF
--- a/internal/database/strikes.go
+++ b/internal/database/strikes.go
@@ -40,11 +40,10 @@ type Strike struct {
 	StrikeID       int
 	Reason         string
 	DiscordUserID  string
-	DiscordGuildID string
 }
 
 func ListUserStrikes(conn *pgxpool.Pool, userID int) ([]Strike, error) {
-	query := `SELECT s.strikeID, s.reason, u.discordUserID, u.discordGuildID
+	query := `SELECT s.strikeID, s.reason, u.discordUserID
 	FROM strikes s
 	JOIN users u ON s.userID = u.userID
 	WHERE s.userID = $1;`
@@ -62,7 +61,6 @@ func ListUserStrikes(conn *pgxpool.Pool, userID int) ([]Strike, error) {
 			&currRow.StrikeID,
 			&currRow.Reason,
 			&currRow.DiscordUserID,
-			&currRow.DiscordGuildID,
 		)
 		if err != nil {
 			return nil, err
--- a/internal/database/strikes.go
+++ b/internal/database/strikes.go
@@ -1,0 +1,98 @@
+package database
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+type AddStrikeEntryArgs struct {
+	DbUserID   int
+	Reason     string
+	StrikeTime string
+}
+
+func AddStrike(conn *pgxpool.Pool, p AddStrikeEntryArgs) (int, int, error) {
+	query := `INSERT INTO strikes (userID, reason, createTimestamp)
+	VALUES ($1, $2, $3)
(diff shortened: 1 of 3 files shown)


internal/database/strikes.go
```diff
@@ -0,0 +1,98 @@
...
+	// currently added strike is not counted when current strike count is returned
+	// Adding one here makes sure correct value is returned
+	return strikeID, strikeCount + 1, nil
+}
+
+type Strike struct {
+	StrikeID       int
+	Reason         string
+	DiscordUserID  string
+	DiscordGuildID string
```
brtran4, line 43: for now we can just use the constant `GuildID` as we're not planning on releasing this bot to other discord servers

brtran4 (commented):
can you include a screenshot of what the embed looks like?

internal/database/strikes.go
KingPieson: This just pulls from the user table, so I don't think it should ever be a different value, and since we already need to query users it doesn't add any real extra amount of overhead to just pull that info directly

KingPieson:
> can you include a screenshot of what the embed looks like?

I'm not quite sure what the question here is, sorry.

internal/database/strikes.go
brtran4: eh i guess it's fine, but i still don't really think we need a field for that since we're probably never going to extend this bot outside of NAUR

brtran4:
> > can you include a screenshot of what the embed looks like?
> 
> I'm not quite sure what the question here is, sorry.

there should be an embedded message sent to the mod logging channel that displays the action such as (x admin used /strike on x user). it should look something like this, but for the strikes system

Screenshot 2024-09-09 at 12 53 29 PM

internal/database/strikes.go
KingPieson: fair, this is just printing anyways so it's easy to remove, done

KingPieson:
image
image

This PR is not enabling any of the `/strike` commands, just setting up DB interactions

internal/discord/commands.go
```diff
@@ -272,7 +272,53 @@ func (d *Discord) Unexile(s *discordgo.Session, i *discordgo.InteractionCreate)
...
+	strikeID, strikeCount, err := database.AddStrike(d.Conn, strikeEntryArgs)
+	if err != nil {
+		tempstr := "Unable to add entry to the database"
+		log.Printf("%v: %v\n", tempstr, err)
+		RespondAndAppendLog(state, tempstr)
+		return
+	}
+
+	tempstr := fmt.Sprintf(
+		"User <@%v> has been given a strike. This is strike number %v",
```
brtran4, line 314: @clairefleuret do we need to state how many strikes this person receives when it's going to be logged in the mod actions channel anyways?

brtran4: ok then let's add a period at the end of the number

KingPieson: Added

internal/discord/commands.go
```diff
@@ -272,7 +272,53 @@ func (d *Discord) Unexile(s *discordgo.Session, i *discordgo.InteractionCreate)
...
+	logMsg, _ := d.LogCommand(i.Interaction)
+
+	state := &InteractionState{
+		session:     s,
+		interaction: i,
+		logMsg:      logMsg,
+		isFirst:     true,
+	}
+
+	userToExile := optionMap["user"].UserValue(nil)
```
brtran4, line 285: probably should change this to `userToStrike`

KingPieson: whoops copy paste mistake. fixed

brtran4 (approved):

PR #106, opened 2024-09-28
Initial python setup

KingPieson:
Initial python rewrite skeleton. Extra work that needs to get done is listed in main.py

### DIFF
--- a/README-python.md
+++ b/README-python.md
@@ -1,0 +1,32 @@
+# Moddingway
+
+Discord moderation bot for NAUR.
+
+### Environment variables
+Postgres-related information is configured in the environment variables instead of a pre-created user/password.
+#### Release
+- DISCORD_TOKEN
+- POSTGRES_HOST
+- POSTGRES_PORT
+- POSTGRES_DB
+- POSTGRES_USER
+- POSTGRES_PASSWORD
+
+#### Testing
+- DISCORD_TOKEN
+- POSTGRES_HOST
+- POSTGRES_PORT
+- POSTGRES_DB
+- POSTGRES_USER
+- POSTGRES_PASSWORD
+- DEBUG
+- GUILD_ID
+- MOD_LOGGING_CHANNEL_ID
+
+`DEBUG` must be set to `true` for testing.  
+Defaults are also set for `POSTGRES_PORT` (5432) and `POSTGRES_DB` (moddingway) if those two are not set.
+
+## Development recommendations
+
+#### Black Formatter
+Files in this repo will be run through the [Black Formatter](https://black.readthedocs.io/en/stable/). To minimize merge conflicts, it is recommended to run this formatter on your code before submitting. Most IDEs will have an extension for black, and it is recommended to use those
--- a/src/bot.py
+++ b/src/bot.py
@@ -1,7 +1,7 @@
 import discord
 from discord.ext.commands import Bot
 from settings import get_settings
-from commands.exile import create_exile_command, create_unexile_command
+from commands.exile import create_exile_commands
 import logging
 
 settings = get_settings()
@@ -10,10 +10,7 @@ logger = logging.getLogger(__name__)
 
 class ModdingwayBot(Bot):
     async def setup_hook(self):
-        logger.debug("setting up commands")
-        create_unexile_command(self)
-        create_exile_command(self)
-        logger.debug("finished setting up commands")
+        self._register_commands()
 
         guild = discord.Object(id=settings.guild_id)
         self.tree.copy_global_to(guild=guild)
@@ -21,3 +18,6 @@ class ModdingwayBot(Bot):
     
     async def on_ready(self):
         logger.info(f'Logged in as {self.user} (ID: {self.user.id})')
+    
+    def _register_commands(self):
+        create_exile_commands(self)
--- a/src/bot.py
+++ b/src/bot.py
@@ -15,9 +15,9 @@ class ModdingwayBot(Bot):
         guild = discord.Object(id=settings.guild_id)
         self.tree.copy_global_to(guild=guild)
         await self.tree.sync(guild=guild)
-    
+
     async def on_ready(self):
-        logger.info(f'Logged in as {self.user} (ID: {self.user.id})')
-    
+        logger.info(f"Logged in as {self.user} (ID: {self.user.id})")
+
     def _register_commands(self):
         create_exile_commands(self)
--- a/src/bot.py
+++ b/src/bot.py
@@ -1,0 +1,23 @@
+import discord
+from discord.ext.commands import Bot
+from settings import get_settings
+from commands.exile import create_exile_command, create_unexile_command
+import logging
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+class ModdingwayBot(Bot):
+    async def setup_hook(self):
+        logger.debug("setting up commands")
+        create_unexile_command(self)
+        create_exile_command(self)
+        logger.debug("finished setting up commands")
+
+        guild = discord.Object(id=settings.guild_id)
+        self.tree.copy_global_to(guild=guild)
+        await self.tree.sync(guild=guild)
+    
+    async def on_ready(self):
+        logger.info(f'Logged in as {self.user} (ID: {self.user.id})')
--- a/src/bot.py
+++ b/src/bot.py
@@ -1,7 +1,7 @@
 import discord
 from discord.ext.commands import Bot
 from settings import get_settings
-from commands.exile import create_exile_commands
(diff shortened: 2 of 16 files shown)


src/bot.py
```diff
@@ -0,0 +1,23 @@
...
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+class ModdingwayBot(Bot):
+    async def setup_hook(self):
+        logger.debug("setting up commands")
+        create_unexile_command(self)
+        create_exile_command(self)
```
brtran4, line 15: i think we could shove this in a helper function so that we don't have to revisit this function every time we add a new command

src/util.py
```diff
@@ -0,0 +1,5 @@
+import discord
+
+def log_info_and_embed(embed: discord.Embed, logger, message: str):
+    embed.description += "\n" +  message
+    logger.info(message)
```
brtran4, line 5: im thinking we also add a TODO list here to send DMs to the user

src/commands/exile.py
```diff
@@ -0,0 +1,53 @@
...
+from services.exile_service import exile_user, unexile_user
+from collections.abc import Coroutine
+
+settings = get_settings()
+
+def create_unexile_command(bot: Bot) -> None:
+    @bot.tree.command()
+    @discord.app_commands.describe(
+        user="User being exiled"
+    )
```
brtran4, line 13: i think it should be legal to put this above the function signature. that way we won't have to create a nested function

src/commands/exile.py
```diff
@@ -0,0 +1,53 @@
...
+        """Exile the specified user."""
+
+        await run_command_with_logging(interaction, exile_user, user, duration, reason)
+
+        await interaction.response.send_message(f"Successfully exiled {user.mention}", ephemeral=True)
+
+
+# NB this feels very messy, fragile, and easy to mess up
+# There might be a better way to wrap all this in something cleaner and better reusable
+async def run_command_with_logging(interaction, command: Coroutine, *args):
```
brtran4, line 38: not sure if we would also want to write to the logging buffer in this function as well. imo we should use `log_info_and_embed` for the error messages/handling and then use `run_command_with_logging` for the successful operations

brtran4 (changes requested):
just wanted to clarify, what's the difference between exile.py and exile_service.py? is this where `*.py` is the command setup and `*_service.py` is the actual operation?

KingPieson:
> what's the difference between exile.py and exile_service.py
I think naming can be cleaned up here, the intent is to have the methods in the `commands` directory be code that directly interfaces with discord, and the code in the `services` directory be code that actually holds the business logic. This can help once we do things like strikes, where we need to take some action and also exile a user, we will have the code to exile a user be easy to call from multiple places

src/bot.py
KingPieson: I pulled this into a method outside the setup hooks event, and combined the two exile command

src/util.py
KingPieson: The intent with this method is to combine writing to logs and to the discord message we will send to the logging channel in one place. While we will need to be able to send a DM to the user, we won't always want to do that when we're logging things

src/commands/exile.py
KingPieson: I have the async method nested inside the "create command" method in order to move the creation of commands to their own files, while also being able to pass around the `Bot` instance properly. I think there's a way to simplify this using the `ApplicationCommand` class that discord.py exposes, but I was having some trouble in my first pass of that. I agree that this setup is confusing, but I got the pattern from how [mooncord defines its commands](https://github.com/Shy-16/Mooncord-PitBot/blob/master/application_commands/pit_command.py#L9-L20)

src/commands/exile.py
KingPieson: I've updated `log_info_and_embed` and `run_command_with_logging` to try and make it a bit more clear the purposes of each, the intent is to have `log_info_and_embed` be run any time you want to write a single log message, and `run_command_with_logging` be a wrapper around the command's business logic to handle the creation of the embed, and the sending of that embed to the logging discord channel for you

src/util.py
brtran4: er mb i shouldve been clear, i meant at the end of the file to add a dm function

KingPieson: Yeah, sending DMs makes sense to live in the utils file unless it needs a lot of weird interactions with discord objects

brtran4: I think it shouldnt need more than one since we should be able to just pass the discord user id string over and have discord process that on its side

README-python.md
```diff
@@ -0,0 +1,32 @@
...
+- GUILD_ID
+- MOD_LOGGING_CHANNEL_ID
+
+`DEBUG` must be set to `true` for testing.  
+Defaults are also set for `POSTGRES_PORT` (5432) and `POSTGRES_DB` (moddingway) if those two are not set.
+
+## Development recommendations
+
+#### Black Formatter
+Files in this repo will be run through the [Black Formatter](https://black.readthedocs.io/en/stable/). To minimize merge conflicts, it is recommended to run this formatter on your code before submitting. Most IDEs will have an extension for black, and it is recommended to use those
```
brtran4, line 32: we should probably make the github pipeline use black as well

src/settings.py
```diff
@@ -0,0 +1,39 @@
+from pydantic import BaseModel
+import os
+import sys
+import logging
+
+
+class Settings(BaseModel):
+    """Class for keeping track of settings"""
+
+    guild_id: int
```
brtran4, line 10: might want to double check, but i was under the assumption that the guild id is a string

src/settings.py
```diff
@@ -0,0 +1,39 @@
...
+    discord_token: str = os.environ["DISCORD_TOKEN"]
+    log_level: int = logging.INFO
+    logging_channel_id: int
```
brtran4, line 13: same here

src/settings.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def prod() -> Settings:
+    return Settings(
+        guild_id=1172230157776466050,
+        logging_channel_id=int(os.environ["MOD_LOGGING_CHANNEL_ID"]),
```
brtran4, line 19: this should be a fixed channel ID as well. you can get this info from @clairefleuret

src/util.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+async def send_dm(member: discord.Member, messageContent: str):
+    channel = await member.create_dm()
+    await channel.send(content=messageContent)
+
+
+# NB this feels very messy, fragile, and easy to mess up
+# There might be a better way to wrap all this in something cleaner
```
brtran4, line 27: i think this is the best we'll get with this

src/util.py
```diff
@@ -0,0 +1,78 @@
...
+
+async def add_and_remove_role(
+    member: discord.Member, role_to_add: enums.Role, role_to_remove: enums.Role
+):
+    discord_role_to_add = next(
+        (role for role in member.guild.roles if role.name == role_to_add.value), None
+    )
+    discord_role_to_remove = next(
+        (role for role in member.guild.roles if role.name == role_to_remove.value), None
+    )
```
brtran4, line 60: i think in terms of run time it would be faster to do something like the below. i dont think it's worth it to iterate through the roles twice

```suggestion
discord_role_to_add = None
discord_role_to_remove = None
for role in member.guild.roles:
   if role.name == role_to_add.value:
       discord_role_to_add = role
   if role.name == role_to_remove.value:
       discord_role_to_remove = role
```

brtran4 (changes requested):

README-python.md
KingPieson: Agreed, I'll look into pipelines and whatnot when I'm getting to dockerization

src/settings.py
KingPieson: The guild IDs I've seen when doing devwork have all been numeric, and the discord.py `Guild` class has id as an int, so I believe it's numeric

src/util.py
KingPieson: I've updated to run this in a single pass, but I also suspect that our runtimes and numbers of roles are not going to be particularly high or problematic. DB interactions are probably going to be our primary bottleneck

brtran4: yeah theres definitely no getting around that, but it saves us some unnecessary computation for a worst case scenario that we might run into semi-often

brtran4 (approved):
lgtm except for the modlogging channel

src/util.py
KingPieson: I updated this to use the context manager pattern, it's now way more straightforward

brtran4:
since #107 went in we can probably close this PR

PR #109, opened 2024-10-07
Set up DB connection, implement DB aspects of exile/unexile commands

KingPieson:
Set up DB connection on bot spin up. Create DB files to run methods, create exile and unexile records when commands are run

### DIFF
--- a/src/services/exile_service.py
+++ b/src/services/exile_service.py
@@ -4,12 +4,17 @@ from util import log_info_and_embed
 
 logger = logging.getLogger(__name__)
 
-async def exile_user(logging_embed: discord.Embed, user: discord.User, duration: str, reason: str):
+
+async def exile_user(
+    logging_embed: discord.Embed, user: discord.User, duration: str, reason: str
+):
     # TODO implement this functionality
-    log_info_and_embed(logging_embed, logger, f"going to exile {user.mention} for {duration}")
+    log_info_and_embed(
+        logging_embed, logger, f"going to exile {user.mention} for {duration}"
+    )
 
     log_info_and_embed(logging_embed, logger, f"because {reason}")
-    
+
     # look up user in DB
 
     # add exile entry into DB
@@ -18,6 +23,7 @@ async def exile_user(logging_embed: discord.Embed, user: discord.User, duration:
 
     # message user?
     log_info_and_embed(logging_embed, logger, "Job's done")
+
 
 async def unexile_user(logging_embed: discord.Embed, user: discord.User):
     pass
--- a/src/services/exile_service.py
+++ b/src/services/exile_service.py
@@ -15,19 +15,15 @@ async def exile_user(
     user: discord.Member,
     duration: Optional[datetime.timedelta],
     reason: str,
-):
-    # change user role
-    await add_and_remove_role(
-        user, role_to_add=Role.EXILED, role_to_remove=Role.VERIFIED
-    )
-
-    # message user
-    await send_dm(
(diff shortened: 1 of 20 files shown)


src/services/exile_service.py
```diff
@@ -1,34 +1,81 @@
...
+        user, role_to_add=Role.EXILED, role_to_remove=Role.VERIFIED
     )
 
-    # Look up 
-
-    log_info_and_embed(logging_embed, logger, f"because {reason}")
+    # message user
+    await send_dm(
+        user, f"You are being exiled from NAUR FFXIV for the following reason: {reason}"
+    )
```
brtran4, line 27: i think we should move this near the end after all the db interactions are completed

src/services/exile_service.py
```diff
@@ -1,34 +1,81 @@
...
-    await add_and_remove_role(user, role_to_add=Role.EXILED, role_to_remove=Role.VERIFIED)
+    # add exile entry into DB
+    start_timestamp = datetime.datetime.now(datetime.timezone.utc)
+    end_timestamp = None
+    exile_status = ExileStatus.INDEFINITE_EXILE
+    if duration:
+        end_timestamp = start_timestamp + duration
+        exile_status = ExileStatus.TIMED_EXILED
 
+    exile = Exile(None, db_user.user_id, reason, exile_status.value, start_timestamp, end_timestamp)
```
brtran4, line 50: is there a reason why the class will contain an exile_id if it's not actually holding that information?

src/services/exile_service.py
```diff
@@ -1,34 +1,81 @@
...
+    logging_embed: discord.Embed,
+    user: discord.Member,
+    duration: Optional[datetime.timedelta],
+    reason: str,
 ):
-    # TODO implement this functionality
-    log_info_and_embed(
-        logging_embed, logger, f"going to exile {user.mention} for {duration}"
+    # change user role
+    await add_and_remove_role(
```
brtran4, line 20: looking at the function definition i think we'll have a bug where the user could potentially not have the role, but we're still attempting to remove it. this is a hard constraint on exile where the user needs the verified role

brtran4 (changes requested):

src/services/exile_service.py
KingPieson: I think I had just misread a request that Claire had about unexile, I agree we should move this to then end, updated

src/services/exile_service.py
KingPieson: I think I had just misread a request that Claire had about unexile, I agree we should move this to then end, updated

src/services/exile_service.py
KingPieson: Makes sense, I've added a check to the service and updated the command to better pass around errors

src/services/exile_service.py
KingPieson: I still want to think about if there's a better way to pass around data without bloating the number of classes, but the idea is that we want postgres to take care of generating the actual PK of the row, so it's set up as null when it's first being created

brtran4: that makes sense, i'm just not sure if there's a situation where we will need to throw around the exile id around except for getting the list of exiles. it's not detailed in the spec right now so i'm not sure if we're actually going to be doing that

src/services/exile_service.py
brtran4: i think we deleted this line on accident

KingPieson: That's just a merge conflict thing, there was two calls to each of those and now they exist at the bottom of the method

brtran4 (approved):

PR #115, opened 2024-10-10
Create reusable embed function

OTCompa:
Adds a base function to create embeds for reusability. Updates the existing `create_logging_embed()` to use the function and adds command parameters to the embed. Additionally sets the footer to the relevant exile ID in applicable functions.
Base function will be reused with auto unexile

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -4,7 +4,7 @@ from settings import get_settings
 from services.exile_service import exile_user, unexile_user
 from util import is_user_moderator, calculate_time_delta
 from typing import Optional
-from .helper import create_logging_embed, get_args
+from .helper import create_logging_embed
 
 settings = get_settings()
 
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -4,7 +4,7 @@ from settings import get_settings
 from services.exile_service import exile_user, unexile_user
 from util import is_user_moderator, calculate_time_delta
 from typing import Optional
-from .helper import create_logging_embed
+from .helper import create_logging_embed, get_args
 
 settings = get_settings()
 
@@ -16,7 +16,7 @@ def create_exile_commands(bot: Bot) -> None:
     async def unexile(interaction: discord.Interaction, user: discord.Member):
         """Unexile the specified user."""
 
-        async with create_logging_embed(interaction) as logging_embed:
+        async with create_logging_embed(interaction, get_args()) as logging_embed:
             error_message = await unexile_user(logging_embed, user)
 
             await interaction.response.send_message(
@@ -37,8 +37,7 @@ def create_exile_commands(bot: Bot) -> None:
         reason: str,
     ):
         """Exile the specified user."""
-
-        async with create_logging_embed(interaction) as logging_embed:
+        async with create_logging_embed(interaction, get_args()) as logging_embed:
             exile_duration = calculate_time_delta(duration)
             if duration and not exile_duration:
                 await interaction.response.send_message(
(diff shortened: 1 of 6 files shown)


src/util.py
```diff
@@ -8,6 +9,49 @@
...
+        if fields is not None:
+            for field in fields:
+                embed.add_field(name=field.name, value=field.value, inline=False)
+
+        yield embed
+    except Exception as e:
+        embed.add_field(name="Error", value=e, inline=False)
+        raise e
+    finally:
+        await log_channel.send(embed=embed)
```
brtran4, line 52: so the function name implies that we are only creating the embed and not sending it to the logging channel, which is two operations. so we can choose between two options
1. rename the function to be `create_and_send_embed`
2. return the embed here and leave the sending to the caller

KingPieson: My 2 cents on a better name is `logging_embed_context` , since it's a context manager for the logging embed. Since every time we use this, we're going to want to be sending the embed to the logging channel, I'd rather not go with option 2

src/commands/helper.py
```diff
@@ -1,27 +1,36 @@
...
-    log_channel = interaction.guild.get_channel(settings.logging_channel_id)
-    try:
-        embed.set_author(
-            name=interaction.user.display_name,
-            icon_url=interaction.user.display_avatar.url,
-        )
-        embed.timestamp = interaction.created_at
-        embed.description = f"Used `{interaction.command.name}` command in {interaction.channel.mention}"
-        embed.add_field(name="Action", value=f"/{interaction.command.name}")
+def create_logging_embed(interaction: discord.Interaction, args: Optional[dict] = None):
```
KingPieson, line 10: I think the way these methods are set up should be flipped. The commands should be interacting directly with the context manager via `async with logging_embed_context(...) as logging_embed:` and then `logging_embed_context` should be calling `create_logging_embed` to build the initial embed

src/commands/helper.py
```diff
@@ -1,27 +1,36 @@
...
-        raise e
-    finally:
-        await log_channel.send(embed=embed)
+
+def get_args():
+    """Returns all arguments in the calling function except the first one (interaction)"""
+    calling_frame = sys._getframe().f_back
+    caller = calling_frame.f_code
+    relevant_vars = caller.co_varnames[: caller.co_argcount]
+    return {name: calling_frame.f_locals[name] for name in relevant_vars[1:]}
```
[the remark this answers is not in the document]
KingPieson: I agree that passing around the method arguments and making these sorts of system calls feels like it's probably much more complicated than it needs to be. I think the better way to approach this is to have the method that creates the context manager take in kwargs and then the method creating the manager can populate it with whatever data is expected. If that's not clear, I can toss in some example code that might make things easier

src/commands/helper.py
OTCompa: I'm afraid I don't follow. I intended to reuse `logging_embed_context(...)` in the autounexile log like so:
```python
def create_autounexile_embed(
    self, user: discord.User, exile_id: str, end_timestamp: str
):
    return logging_embed_context(
        self.get_channel(settings.logging_channel_id),
        user=user,
        timestamp=end_timestamp,
        description=f"<@{user.id}>'s exile has timed out",
        footer=f"Exile ID: {exile_id}",
    )
```

KingPieson: Ah okay I see, that makes sense. Then let's rename this a little bit to still make it clear we're creating a context, something like `create_interaction_embed_context` and then the autounexile could be `create_autounexile_embed_context`

KingPieson (approved):

src/commands/exile_commands.py
```diff
@@ -4,7 +4,7 @@
 from services.exile_service import exile_user, unexile_user
 from util import is_user_moderator, calculate_time_delta
 from typing import Optional
-from .helper import create_logging_embed
+from .helper import create_logging_embed, get_args
```
brtran4, line 7: looks like we forgot to remove the `get_args` import

brtran4 (approved):

PR #116, opened 2024-10-11
Python app docker container

KingPieson:
Add python app dockerfile and update the python readme a little to make onboarding easier


src/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.11
+
+# install requiremed packages
```
brtran4, line 3: nit: requiremed -> required

brtran4 (approved):

Makefile
```diff
@@ -22,4 +22,10 @@ test-build:
 	docker compose -f postgres.yml down
 	docker compose -f postgres.yml up --build
 
+python-docker-build:
+	docker-compose build python-app
```
brtran4, line 26: iirc it should be `docker compose`

docker-compose.yml
```diff
@@ -1,3 +1,7 @@
+include:
+  - path: 
+      - postgres.yml
+
```
OTCompa, line 4: AFAIK the postgres container should only run in dev environments, so any reference to the postgres container shouldn't be in `docker-compose.yml`.
in prod, atmus has a separate database that we will be using, so the base container shouldn't have a postgres database. 

when I initially set the docker containers up, the idea was that the default docker compose only runs the app container. if someone wanted to start a dev environment with the postgres container, running `make test-build` would run `postgres.yml`, which includes `docker-compose.yml` and `postgres.override.yml`. `postgres.override.yml` would override some fields in `docker-compose.yml` to make the app container dependent on the postgres container and set the `POSTGRES_HOST` variable to the proper service name. my explanation may be confusing so here's what it should look like: https://github.com/naurffxiv/moddingway/compare/python_app_docker_container...docker-example

in hindsight I should've commented what each file was for to make the intention clearer

OTCompa (changes requested):

Makefile
KingPieson: yeah sorry, I'm used to the older version of docker compose, we should be using the newer `docker compose` instead where we can

docker-compose.yml
KingPieson: Yeah good call about the prod version, I'll remove the DB dependency from the app docker compose. However, I think having the database container automatically spin up the app containers is a bit confusing. I think it makes more sense to have the postgres pod and the app pod both be independent, then have the following make commands (names are placeholder)

1. make database (this spins up postgres container only)
2. make app (spins up both the app and DB pods)

Do you know if having things configured that way would interface nicely with the overrides you have set up, or do we need to have the postgres file spin up the app in order to get that to work?

OTCompa: It should interface nicely by just moving the include and renaming the override. However I think you'll need to set up a network between both containers since they're now two completely separate containers instead of services within the same container. Another minor thing is that the app container won't be able to wait on the database container but I assume that can be dealt with in code if it's really that big of an issue

KingPieson: I pushed out a change for this, I ended up making another service in the main docker-compose file that extended the base python container. This let me properly have the container wait for the DB to be healthy before starting and also let me easily override the db host to point to the other container. I didn't need to mess with the docker environment stuff at all, but it seems to be working on my side. If anyone runs into issues getting things to run feel free to tag me and I can take a look

Makefile
```diff
@@ -22,4 +22,14 @@ test-build:
 	docker compose -f postgres.yml down
 	docker compose -f postgres.yml up --build
 
+python-build:
+	docker compose build python-app
+
+python-run:
+	docker compose down
+	docker compose run python-app-local
```
OTCompa, line 30: nit: use `docker compose up` instead of `docker compose run`

docker-compose.yml
```diff
@@ -29,3 +33,27 @@ services:
...
+      MOD_LOGGING_CHANNEL_ID: ${MOD_LOGGING_CHANNEL_ID}
+
+  python-app-local:
+      extends:
+        service: python-app
+      environment:
+        POSTGRES_HOST: postgres_local
+      depends_on:
+        postgres_local:
+          condition: service_healthy
```
OTCompa, line 59: nit: indent isn't consistent with the rest of the file

docker-compose.yml
OTCompa: just tested both with and without the postgres db, didn't run into any issues!

OTCompa (approved):

PR #117, opened 2024-10-15
Add roulette command to exile user with random duration

brtran4:


brtran4:
requesting @clairefleuret for text review

clairefleuret (approved):

OTCompa (approved):

src/commands/exile_commands.py
```diff
@@ -54,3 +57,53 @@ async def exile(
...
+                        ephemeral=False,
+                    )
+            else:
+                await interaction.response.send_message(
+                    f"<@{interaction.user.id}> has tested their luck and lives another day...",
+                    ephemeral=False,
+                )
+
+    @bot.tree.error
+    async def on_app_command_error(interaction: discord.Interaction, error):
```
KingPieson, line 97: Nit: this probably should be moved to its own file instead of being combined with the exile commands, but we can move it later

KingPieson (approved):

src/commands/exile_commands.py
brtran4: handled in https://github.com/naurffxiv/moddingway/pull/123

PR #120, opened 2024-10-16
show exile and ban commands via app

KingPieson:
Allow access to ban and exile commands via the app commands interface. Surface both at the user level and at the message level

#### Example flow of exiling user via app command
image

image

image

Ticket: MWAY-17


OTCompa (approved):
Messed around with it and looks good to me. I know normal users can't use it due to the check, but I think it's probably worth mentioning to the others that these are visible to everybody in the server so nobody panics.

src/ui/ban_modal.py
```diff
@@ -0,0 +1,31 @@
...
+class BanModal(discord.ui.Modal):
+
+    def __init__(self, user: discord.Member) -> None:
+        self.user = user
+        super().__init__(title=f"Ban User {user.display_name}")
+
+    reason = discord.ui.TextInput(
+        label="Ban Reason",
+        style=discord.TextStyle.long,
+        placeholder="Why is the user banned",
```
brtran4, line 15: i think we can change this to "Reason for banning user"

src/commands/exile_commands.py
```diff
@@ -93,3 +94,33 @@ async def roulette(interaction: discord.Interaction):
                     f"<@{interaction.user.id}> has tested their luck and lives another day...",
                     ephemeral=False,
                 )
+
+    @bot.tree.context_menu(name="Exile User for custom duration")
+    @discord.app_commands.check(is_user_moderator)
+    async def exile_user_action(interaction: discord.Interaction, user: discord.Member):
+        """Exile the selected user"""
+        await interaction.response.send_modal(ExileModal(user))
```
brtran4, line 102: will we hit the interaction timeout constraint if the DM hangs on us?

src/ui/exile_modal.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self, user: discord.Member) -> None:
+        super().__init__(title=f"Exile User {user.display_name}")
+        self.user = user
+
+    duration = discord.ui.TextInput(label="Exile Duration")
+
+    reason = discord.ui.TextInput(
+        label="Exile Reason",
+        style=discord.TextStyle.long,
+        placeholder="Why is the user exiled",
```
brtran4, line 17: same thing here, "Reason for exiling user"

src/ui/exile_modal.py
```diff
@@ -0,0 +1,47 @@
...
+        required=True,
+    )
+
+    async def on_submit(self, interaction: discord.Interaction):
+
+        async with create_modal_embed(
+            interaction,
+            "Exile User",
+            user=self.user,
+            duration=self.duration.value,
```
brtran4, line 28: we should be safe and error check duration since they are required now

brtran4 (changes requested):

src/commands/exile_commands.py
KingPieson: The interaction doesn't seem to start until you press submit on the modal, so there's plenty of time to fill the form out. Possibly there's a timeout if you leave the modal up for the full discord timeout (I think it's 15 minutes?). And the DM just running a bit long is solved by having these all use the initial resp that we set up previously

src/ui/ban_modal.py
KingPieson: Agreed, done

src/ui/exile_modal.py
KingPieson: Done

src/ui/exile_modal.py
KingPieson: I've marked the field as required on the modal, which has good UI checks and won't let you submit if either field is blank

src/ui/exile_modal.py
```diff
@@ -0,0 +1,58 @@
...
+        placeholder="Reason for exiling user",
+        max_length=512,
+        required=True,
+    )
+
+    async def on_submit(self, interaction: discord.Interaction):
+        exile_duration = calculate_time_delta(self.duration.value)
+
+        if self.duration.value and not exile_duration:
+            # TODO update this string once exile duration change is merged
```
brtran4, line 27: let's add the actual ticket number here

brtran4 (approved):

src/ui/exile_modal.py
KingPieson: I'm just gonna hold off on this merge until after the input formatting change goes in. I rewrote this to have the values all over the place and it's not worth the headache of merging this then breaking it again

PR #128, opened 2024-10-28
Create a command to show user exiles

KingPieson:
* Create command to show user's exiles
* When parsing duration fails on exile command, clearly state that the user will not be exiled
* Do not create log when parsing exile duration fails
* Try/catch around sending user DMs, since failures are expected when user blocks DMs

Screenshot is taken on a 1080p monitor with both discord sidebars open, not sure if there's a better way to handle long reason messages. Lorem ipsum example is 508 characters long

image

image

image

Ticket: MWAY-13 MWAY-4

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -2,7 +2,7 @@ import discord
 import logging
 from discord.ext.commands import Bot
 from settings import get_settings
-from services.exile_service import exile_user, unexile_user
+from services.exile_service import exile_user, unexile_user, get_user_exiles
 from util import is_user_moderator, calculate_time_delta
 from typing import Optional
 from .helper import create_logging_embed
@@ -93,3 +93,16 @@ def create_exile_commands(bot: Bot) -> None:
                     f"<@{interaction.user.id}> has tested their luck and lives another day...",
                     ephemeral=False,
                 )
+
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    @discord.app_commands.describe(user="User whose exile is being checked")
+    async def check_exile(interaction: discord.Interaction, user: discord.Member):
+        """Check the exile status of a user."""
+
+        async with create_logging_embed(interaction, user=user) as logging_embed:
+            msg = await get_user_exiles(logging_embed, user)
+
+            await interaction.response.send_message(
+                msg, ephemeral=True
+            )
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -3,7 +3,7 @@ import logging
 from discord.ext.commands import Bot
 from settings import get_settings
 from services.exile_service import exile_user, unexile_user, get_user_exiles
-from util import is_user_moderator, calculate_time_delta, log_info_and_embed
+from util import is_user_moderator, calculate_time_delta
 from typing import Optional
 from .helper import create_logging_embed
 from random import choice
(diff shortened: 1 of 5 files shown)


src/services/exile_service.py
```diff
@@ -104,3 +110,25 @@ async def unexile_user(
...
+    if db_user is None:
+        return "User not found in database"
+
+    exile_list = exiles_database.get_user_exiles(db_user.user_id)
+
+    if len(exile_list) == 0:
+        return "No exiles found for user"
+
+    # TODO convert this to an object for better stringifying
+    res = "Exiles found for the given user:\n"
```
brtran4, line 126: we should spit out the user id in `@<user_id>` format here for clarity.

src/commands/exile_commands.py
```diff
@@ -93,3 +94,14 @@ async def roulette(interaction: discord.Interaction):
...
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    @discord.app_commands.describe(user="User whose exile is being checked")
+    async def check_exile(interaction: discord.Interaction, user: discord.Member):
+        """Check the exile status of a user."""
+
+        async with create_logging_embed(interaction, user=user) as logging_embed:
+            msg = await get_user_exiles(logging_embed, user)
+
+            await interaction.response.send_message(msg, ephemeral=True)
```
brtran4, line 107: do we want this to be an ephemeral or an actual embedded message? @clairefleuret

src/services/exile_service.py
```diff
@@ -104,3 +110,25 @@ async def unexile_user(
...
+
+    for exile in exile_list:
+        res = (
+            res
+            + f"ID: `{exile[0]}`| Reason: `{exile[1]}` | start date: `{exile[2]}` | end date: `{exile[3]}` | status: `{ExileStatus(exile[4]).name}`"
```
brtran4, line 131: seems like https://stackoverflow.com/questions/63565825/how-to-make-data-to-be-shown-in-tabular-form-in-discord-py is a good resource for forming tables

brtran4 (changes requested):

src/commands/exile_commands.py
clairefleuret: @brtran4 to make it as simple as possible, an ephemeral message makes the most sense. i dont think its required to send to the logging channel or report it there

src/services/exile_service.py
KingPieson: updated

src/services/exile_service.py
KingPieson: I updated the formatting to put this into a table. The code to handle longer reasons is a bit messy, we can truncate after a handful of characters if we want. I also think the way we're configured currently, we never expect to have more than 1 exile for a given user, so we possibly can better handle this situation

src/services/exile_service.py
```diff
@@ -104,3 +111,51 @@ async def unexile_user(
...
+    rows = []
+    for exile in exile_list:
+        if len(exile[1]) < 56:
+            rows.append(
+                [
+                    exile[0],
+                    exile[1],
+                    exile[2].strftime(time_format),
+                    exile[3].strftime(time_format) if exile[3] else "Indefinite",
+                    ExileStatus(exile[4]).name,
```
brtran4, line 137: these variables are kinda hard to read, can we initialize these as new vars?

src/services/exile_service.py
```diff
@@ -104,3 +111,51 @@ async def unexile_user(
...
+    db_user = users_database.get_user(user.id)
+    if db_user is None:
+        return "User not found in database"
+
+    exile_list = exiles_database.get_user_exiles(db_user.user_id)
+
+    if len(exile_list) == 0:
+        return "No exiles found for user"
+
+    time_format = "%Y-%m-%d %H:%M:%S"
```
brtran4, line 126: should this be a const?

src/services/exile_service.py
```diff
@@ -104,3 +111,51 @@ async def unexile_user(
...
+                [
+                    exile[0],
+                    exile[1],
+                    exile[2].strftime(time_format),
+                    exile[3].strftime(time_format) if exile[3] else "Indefinite",
+                    ExileStatus(exile[4]).name,
+                ]
+            )
+        else:
+            rows.append(
```
brtran4, line 141: we should probably add a comment on what we're doing to the reason formatting

brtran4 (changes requested):

src/services/exile_service.py
brtran4: yeah since we're not expecting the more than 1 exile situation we should probably clean this up a little bit to assume that

KingPieson: I've updated this to only show the first exile, as well as cleaned up the code and moves some stuff to be demarcated as constants

src/services/exile_service.py
```diff
@@ -104,3 +111,62 @@ async def unexile_user(
...
+            rows.append(["", exile_reason[i : i + MAX_REASON_WIDTH], "", "", ""])
+            i = i + MAX_REASON_WIDTH
+
+    table_output = table2ascii(
+        header=["ID", "Reason", "Start Date", "End Date", "Status"],
+        body=rows,
+        column_widths=[6, 58, 21, 21, 18],
+    )
+
+    return f"Exiles found for the given <@{user.id}>:\n```{table_output}```"
```
brtran4, line 172: nit: `Exiles found for <@{user.id}>`

brtran4 (approved):

PR #131, opened 2024-10-29
Respond with interim message when command is initially received.

OTCompa:
Creates a context manager for the interaction response. Initially sends an interim message and edits the interim message accordingly upon return.
Currently set message is "Processing..."
image

### DIFF
--- a/src/commands/helper.py
+++ b/src/commands/helper.py
@@ -40,18 +40,12 @@ def create_bot_errors(bot: Bot) -> None:
                 f"This command is on cooldown. Please try again in {hours_left} hour(s).",
                 ephemeral=True,
             )
-        #  To be removed since create_response_context() will handle this case
-        # else:
-        #     # Handle other errors if necessary
-        #     await interaction.response.send_message(
-        #         "An error occurred while processing the command.", ephemeral=True
-        #     )
 
 
 @asynccontextmanager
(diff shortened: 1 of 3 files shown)


src/commands/helper.py
```diff
@@ -39,8 +40,35 @@ async def on_app_command_error(interaction: discord.Interaction, error):
...
-            # Handle other errors if necessary
-            await interaction.response.send_message(
-                "An error occurred while processing the command.", ephemeral=True
-            )
+        #  To be removed since create_response_context() will handle this case
+        # else:
+        #     # Handle other errors if necessary
+        #     await interaction.response.send_message(
+        #         "An error occurred while processing the command.", ephemeral=True
+        #     )
```
brtran4, line 48: we can go ahead and just delete this

src/commands/helper.py
```diff
@@ -39,8 +40,35 @@ async def on_app_command_error(interaction: discord.Interaction, error):
...
+
+
+@asynccontextmanager
+async def create_response_context(interaction: discord.Interaction):
+    # Can't yield a string since it's immutable, so create a helper class
+    class responseHelper:
```
brtran4, line 54: nit: classes always start with upper case

brtran4 (changes requested):

brtran4 (approved):

PR #135, opened 2024-10-31
ci: Add workflow to check for ticket number in PR descriptions

brtran4:
Ticket: MWAY-7

### DIFF
--- a/.github/workflows/commit-check.yaml
+++ b/.github/workflows/commit-check.yaml
@@ -1,0 +1,27 @@
+name: Check Ticket Number in Pull Request Description
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check_ticket_in_pr_description:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check pull request description for ticket number
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TICKET_PATTERN="^\s*(Ticket:\s*)?([A-Z]+-[0-9]+|NA)\s*$"
+
+          pr_description=$(curl -s -H "Authorization: token $GITHUB_TOKEN" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }} | jq -r .body)
+
+          if [[ ! "$pr_description" =~ $TICKET_PATTERN ]]; then
+            echo "The pull request description does not contain a valid ticket number or 'Ticket: NA'."
```
KingPieson, line 23: Is this going to block merges that don't have tickets associated with them? I'd prefer this step to be non-blocking personally, but it's not a deal breaker to me

brtran4: Yeah, but there is an option to label it as not attached to a ticket. IMO since we're attaching these PRs to tickets we should do it consistently in case we need to reference the breaking code or if we need it in the future

clairefleuret (approved):

PR #137, opened 2024-11-02
Delete threads that do not meet activity requirements

OTCompa:
Ticket: [MWAY-15](https://naurffxiv.atlassian.net/browse/MWAY-15)
Adds a task that runs once every 24 hours that checks threads for activity requirements. Additionally, the task also checks if the starter message is present. If the thread doesn't meet either requirements, it's deleted and a DM will be sent (commented out for now to deal with backlog present in NAUR) to the original owner. A log will also be made upon each deletion.

In debug mode, it will check for the `INACTIVE_FORUM_CHANNEL_ID` and `INACTIVE_FORUM_DURATION` environment variables.

# Log message
image

# DM
image

[MWAY-15]: https://naurffxiv.atlassian.net/browse/MWAY-15?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -15,6 +15,7 @@ class Settings(BaseModel):
     database_name: str = "moddingway"
     postgres_username: str = os.environ["POSTGRES_USER"]
     postgres_password: str = os.environ["POSTGRES_PASSWORD"]
+    automod_inactivity: dict[int, int]  # key: channel id, value: inactive limit (days)
 
 
 def prod() -> Settings:
@@ -24,16 +25,30 @@ def prod() -> Settings:
         log_level=logging.INFO,
         postgres_host=os.environ.get("POSTGRES_HOST"),
         postgres_port=os.environ.get("POSTGRES_PORT"),
+        automod_inactivity={
+            1273263026744590468: 30,
+            1273261496968810598: 30,
+            1240356145311252615: 15,
+        },
     )
 
 
 def local() -> Settings:
+    inactive_forum_channel_id = os.environ.get("INACTIVE_FORUM_CHANNEL_ID")
+    inactive_forum_duration = os.environ.get("INACTIVE_FORUM_DURATION")
+
+    if inactive_forum_channel_id is not None and inactive_forum_duration is not None:
+        automod_inactivity = {inactive_forum_channel_id: inactive_forum_duration}
+    else:
+        automod_inactivity = {}
+
     return Settings(
         guild_id=int(os.environ["GUILD_ID"]),
         logging_channel_id=int(os.environ["MOD_LOGGING_CHANNEL_ID"]),
         log_level=logging.DEBUG,
         postgres_host=os.environ.get("POSTGRES_HOST", "localhost"),
         postgres_port=os.environ.get("POSTGRES_PORT", "5432"),
+        automod_inactivity=automod_inactivity,
     )
 
 
--- a/src/settings.py
+++ b/src/settings.py
@@ -26,9 +26,11 @@ def prod() -> Settings:
         postgres_host=os.environ.get("POSTGRES_HOST"),
         postgres_port=os.environ.get("POSTGRES_PORT"),
         automod_inactivity={
-            1273263026744590468: 30,
-            1273261496968810598: 30,
-            1240356145311252615: 15,
+            1273263026744590468: 30,  # lfg
+            1273261496968810598: 30,  # lfm
+            1240356145311252615: 30,  # temporary
+            1301166606985990144: 7,  # FRU
+            1300527846468616302: 7,  # scheduled legacy
         },
     )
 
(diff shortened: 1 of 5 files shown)


src/settings.py
```diff
@@ -24,16 +25,30 @@ def prod() -> Settings:
         log_level=logging.INFO,
         postgres_host=os.environ.get("POSTGRES_HOST"),
         postgres_port=os.environ.get("POSTGRES_PORT"),
+        automod_inactivity={
+            1273263026744590468: 30,
+            1273261496968810598: 30,
+            1240356145311252615: 15,
```
brtran4, line 31: can we leave inline comments to label the channel names?

src/settings.py
```diff
@@ -24,16 +25,30 @@ def prod() -> Settings:
...
+        },
     )
 
 
 def local() -> Settings:
+    inactive_forum_channel_id = os.environ.get("INACTIVE_FORUM_CHANNEL_ID")
+    inactive_forum_duration = os.environ.get("INACTIVE_FORUM_DURATION")
```
brtran4, line 38: can we update the README to include these two vars in the local setup

src/workers/forum_automod.py
```diff
@@ -0,0 +1,69 @@
...
+from discord.utils import snowflake_time
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+@tasks.loop(hours=24)
+async def autodelete_threads(self):
+    guild = self.get_guild(settings.guild_id)
+    if guild is None:
```
brtran4, line 18: let's `logger.error` this with an appropriate message

src/workers/forum_automod.py
```diff
@@ -0,0 +1,69 @@
...
+    for channel_id, duration in settings.automod_inactivity.items():
+        channel = guild.get_channel(channel_id)
+        if channel is None:
+            continue
+
+        for thread in channel.threads:
+            last_post = thread.last_message_id
+
+            if thread.flags.pinned:
+                continue
```
brtran4, line 30: let's move this to the very top of the for loop, group `last_post` to be right above the comment of `starter_message`, and add a comment inside saying "skip the for loop if the thread is pinned"

src/workers/forum_automod.py
```diff
@@ -0,0 +1,69 @@
...
+                    thread,
+                    now,
+                ) as automod_embed:
+                    await thread.delete()
+                    log_info_and_embed(
+                        automod_embed, logger, "Thread has been deleted successfully."
+                    )
+
+                    # TODO: uncomment DM portion and sleep when backlog is dealt with
+                    # if thread.owner is not None:
```
brtran4, line 60: i think we won't ever hit this case because there will always be an owner for a thread

brtran4 (changes requested):

src/workers/forum_automod.py
```diff
@@ -0,0 +1,69 @@
...
+            last_post = thread.last_message_id
+
+            if thread.flags.pinned:
+                continue
+
+            # check if starter message was deleted
+            starter_message = None
+            try:
+                starter_message = await thread.fetch_message(thread.id)
+                await asyncio.sleep(0.5)
```
brtran4, line 36: i'm scared that we'll get rate limited if we're doing a request every 0.5 seconds for the overhead

src/workers/forum_automod.py
OTCompa: the `thread.owner` property returns `Optional[discordgo.Member]`. I believe it will return `None` when the person has left the server.

brtran4: uh hold on, discordgo shouldnt be an object in this case, are we looking at the right documentation?

src/workers/forum_automod.py
OTCompa: discord.py will handle the overall rate limit for us, (5 API requests per second is the limit). This sleep is here to prevent the task from blocking the other parts of the bot

brtran4: ok sounds good to me 👍

src/workers/forum_automod.py
OTCompa: sorry, I specified discordgo.Member since that's how I'd refer tot he class in the code

src/workers/forum_automod.py
brtran4: er sorry i meant the comment being inside the if block right before continue

src/workers/forum_automod.py
brtran4: ok that sounds good to me. if we want, we could be more specific in the errors of DMs being turned off vs user is not in the discord server anymore

OTCompa: oh right, I meant to ask this too. should I add a case for that or is letting the embed handle it good enough?
image

OTCompa: I'm just gonna handle it to make things consistent actually

brtran4 (approved):

PR #142, opened 2024-11-02
Fix autounexile looping on fail

OTCompa:
Ticket: MOD-14
This PR fixes a bug where autounexile did not remove exile entries upon failure and now additionally properly handles cases where the user left during the timed exile. It also fixes a minor typo in the ban command.

[MWAY-26]: https://naurffxiv.atlassian.net/browse/MWAY-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/workers/autounexile.py
+++ b/src/workers/autounexile.py
@@ -1,9 +1,9 @@
 from discord.ext import tasks
-from discord.ext.commands import Bot
 from database import exiles_database
 from services.exile_service import unexile_user
 from settings import get_settings
 from .helper import create_autounexile_embed
+from enums import ExileStatus
 import logging
 
 settings = get_settings()
@@ -33,6 +33,6 @@ async def autounexile_users(self):
                 raise Exception(error_message)
         except Exception:
             logger.info(
-                f"Auto Unexile failed, removing exile entry {exile.exile_id}: user {exile.discord_id}"
+                f"Auto Unexile failed, updating exile status of exile {exile.exile_id}, user {exile.discord_id} to unknown"
             )
-            exiles_database.remove_user_exiles(exile.user_id)
+            exiles_database.update_exile_status(exile.exile_id, ExileStatus.UNKNOWN)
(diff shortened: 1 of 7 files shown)


src/workers/autounexile.py
```diff
@@ -16,15 +16,21 @@ async def autounexile_users(self):
...
+                    raise Exception(error_message)
 
-        exiles_database.remove_user_exiles(
-            exile.user_id
-        )  # remove entry from database no matter what
+                error_message = await unexile_user(autounexile_embed, member)
+            if error_message is not None:
+                raise Exception(error_message)
+        except Exception:
+            logger.info(f"Auto Unexile failed, removing exile entry {exile.exile_id}")
```
KingPieson, line 35: Can we also log the user's discordID here? In case moderation wants to manually interact with the user

brtran4: we should also leave a message for mods to make sure that the roles look correct

OTCompa: The embed should already have the user's discord ID but I also added it to the logger just in case!

KingPieson (approved):

src/workers/helper.py
```diff
@@ -2,17 +2,18 @@
...
 from settings import get_settings
 from util import create_interaction_embed_context
+from typing import Optional
 
 settings = get_settings()
 
 
 def create_autounexile_embed(
-    self, user: discord.User, exile_id: str, end_timestamp: str
+    self, user: Optional[discord.Member], discord_id, exile_id: str, end_timestamp: str
```
brtran4, line 11: nit: since all the other vars are typed, we might as well type `discord_id`

brtran4 (changes requested):

brtran4 (approved):
i'm hesitant to press approve because of the new requirement that we don't drop exiles, but i think we should do it in https://naurffxiv.atlassian.net/browse/MOD-40. so mini approve?

clairefleuret:
This should be merged as a temp fix even if we are not going to be dropping exiles in the future.

PR #143, opened 2024-11-02
make roulette safe 5/6 times

KingPieson:
Make Roulette safe 5 times out of 6, and then make it always pick a duration if you fail

Ticket: MOD-18

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -75,7 +75,9 @@ def create_exile_commands(bot: Bot) -> None:
         duration_string = f"{duration_choice}h"
 
         async with create_response_context(interaction, False) as response_message:
-            async with create_logging_embed(interaction, duration=duration_string) as logging_embed:
+            async with create_logging_embed(
+                interaction, duration=duration_string
+            ) as logging_embed:
                 if safety_choice:
                     response_message.set_string(
                         f"<@{interaction.user.id}> has tested their luck and lives another day..."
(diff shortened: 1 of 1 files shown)


OTCompa (approved):

src/commands/exile_commands.py
```diff
@@ -68,14 +68,20 @@ async def exile(
...
             async with create_logging_embed(interaction) as logging_embed:
-                if rand_choice != 0:
+                if safety_choice:
+                    response_message.set_string(
+                        f"<@{interaction.user.id}> has tested their luck and lives another day..."
+                    )
+                else:
                     reason = "roulette"
-                    exile_duration = calculate_time_delta(f"{rand_choice}h")
+                    exile_duration = calculate_time_delta(f"{duration_choice}h")
```
brtran4, line 84: grr it doesnt let me comment on the other file but we should also add the duration to the embed just like we do for exile

brtran4 (changes requested):

src/commands/exile_commands.py
KingPieson: @brtran4 I think I know what you're calling out, I've added it but I think it's a little confusing since the embed is going to have the duration now regardless of if the user was actually exiled or not. If I'm misunderstanding the comment LMK and I can fix

brtran4: yeah i couldnt remember if we already fixed it or not but seems like it's in another PR. lgtm

brtran4 (approved):

PR #144, opened 2024-11-02
Show exile duration formatting in command prompt

KingPieson:
Also remove some logs placed during outage debugging

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -31,7 +31,7 @@ def create_exile_commands(bot: Bot) -> None:
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(
         user="User being exiled",
-        duration="The duration of the exile. TBA format",
+        duration="The duration of the exile. Duration should be in the form of [1 or 2 digits][s, d, m, h]",
         reason="Reason for exile",
     )
     async def exile(
--- a/src/bot.py
+++ b/src/bot.py
@@ -13,11 +13,9 @@ logger = logging.getLogger(__name__)
 
 class ModdingwayBot(Bot):
     async def setup_hook(self):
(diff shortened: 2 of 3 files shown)


OTCompa (approved):

brtran4 (approved):
~tbh it's not a bad thing to keep these~ thought about it a little more and it doesnt really add much

src/commands/exile_commands.py
```diff
@@ -31,7 +31,7 @@ async def unexile(interaction: discord.Interaction, user: discord.Member):
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(
         user="User being exiled",
-        duration="The duration of the exile. TBA format",
+        duration="The duration of the exile. Duration should be in the form of [1 or 2 digits][s, d, m, h]",
```
clairefleuret, line 34: Unless you are very technically inclined, this is awful to read as an end user. My suggestion is: 
`Enter a value Between 1 and 99 followed by s, m, h, or d. Ex: 1s, 1m, 1h, 1d.`

Also, i think we should think about changing the the characters after the numerical value. M can be mistaken for mins or months..

I feel they should be:
- sec
- min
- hour
- day

This would then change the tooltip to:
`Enter a value Between 1 and 99 followed by sec, min, hour, or day. Examples: 1sec, 1min, 1hour, 1day.`

clairefleuret (dismissed):

src/commands/exile_commands.py
KingPieson: @clairefleuret I think this makes a lot of sense, but I'm gonna pull it into its own PR since it's changing actual code rather than just a comment

clairefleuret: If you can make a ticket that would be great.

PR #145, opened 2024-11-02
Condense automod logs

OTCompa:
Condenses automod logs from a per-thread basis to a per-forum channel basis.
# Preview of embed
image
First embed: Failed DM, per person (on failed DM)
Second embed: Total stats, per forum channel

### DIFF
--- a/src/workers/forum_automod.py
+++ b/src/workers/forum_automod.py
@@ -82,7 +82,13 @@ async def autodelete_threads(self):
                 logger.error(e)
                 num_error += 1
             # await asyncio.sleep(300)
-        async with create_automod_embed(
-            self, channel_id, num_removed, num_error, datetime.now(timezone.utc)
-        ):
-            pass
+        if num_removed > 0 or num_error > 0:
+            logger.info(
+                f"Removed a total of {num_removed} threads from channel {channel_id}. {num_error} failed removals."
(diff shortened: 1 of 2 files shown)


clairefleuret:
@OTCompa what would happen if a DM is failed to be sent

OTCompa:
> @OTCompa what would happen if a DM is failed to be sent

It would log it in the program itself, no mention of it in the embed. If you want stats on that I can add it

clairefleuret:
Yes, any failed dm should be logged

clairefleuret (approved):
Looks good to me from user perspective

src/workers/forum_automod.py
```diff
@@ -12,6 +12,10 @@
 logger = logging.getLogger(__name__)
 
 
+class UnableToDM(Exception):
+    pass
+
+
 @tasks.loop(hours=24)
```
KingPieson, line 19: I think this should be at a specified time, rather than just once every 24 hours. Not blocking for now, we can revisit later

KingPieson (approved):

src/workers/forum_automod.py
OTCompa: I'll make an issue on the board for that and figure it out down the line👍

PR #146, opened 2024-11-03
Automod archived threads

OTCompa:
Ticket: MOD-27

[MOD-27]: https://naurffxiv.atlassian.net/browse/MOD-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util.py
+++ b/src/util.py
@@ -19,6 +19,8 @@ class EmbedField(object):
 
 
 class UnableToDM(RuntimeError):
+    # NB: create a custom exception with RuntimeError as base
+    # constructor/methods/attributes are all inherited
     pass
 
 
--- a/src/util.py
+++ b/src/util.py
@@ -18,6 +18,10 @@ class EmbedField(object):
         self.value = value
 
 
+class UnableToDM(RuntimeError):
+    pass
+
+
 @asynccontextmanager
 async def create_interaction_embed_context(
     log_channel: discord.abc.GuildChannel, **kwargs
--- a/src/workers/forum_automod.py
+++ b/src/workers/forum_automod.py
@@ -4,7 +4,7 @@ import logging
 import asyncio
 from settings import get_settings
 from util import send_dm, create_interaction_embed_context
-from .helper import create_automod_embed, automod_thread, AutomodResults
+from .helper import create_automod_embed, automod_thread
 from discord.ext import tasks
 from discord.utils import snowflake_time
 
@@ -20,26 +20,31 @@ async def autodelete_threads(self):
         return
 
     for channel_id, duration in settings.automod_inactivity.items():
-        results = AutomodResults()
+        num_removed = 0
+        num_errors = 0
         channel = guild.get_channel(channel_id)
         if channel is None:
             continue
 
         async for thread in channel.archived_threads(limit=None):
-            await automod_thread(self, channel_id, thread, duration, results)
+            num_removed, num_errors = await automod_thread(
+                self, channel_id, thread, duration, num_removed, num_errors
+            )
 
         for thread in channel.threads:
-            await automod_thread(self, channel_id, thread, duration, results)
+            num_removed, num_errors = await automod_thread(
+                self, channel_id, thread, duration, num_removed, num_errors
+            )
 
-        if results.num_removed > 0 or results.num_errors > 0:
+        if num_removed > 0 or num_errors > 0:
             logger.info(
-                f"Removed a total of {results.num_removed} threads from channel {channel_id}. {results.num_errors} failed removals."
+                f"Removed a total of {num_removed} threads from channel {channel_id}. {num_errors} failed removals."
             )
             async with create_automod_embed(
                 self,
                 channel_id,
-                results.num_removed,
-                results.num_errors,
+                num_removed,
+                num_errors,
                 datetime.now(timezone.utc),
             ):
                 pass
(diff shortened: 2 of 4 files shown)


src/workers/forum_automod.py
```diff
@@ -24,70 +20,27 @@ async def autodelete_threads(self):
...
         channel = guild.get_channel(channel_id)
         if channel is None:
             continue
 
-        for thread in channel.threads:
-            if thread.flags.pinned:
-                # skip the for loop if the thread is pinned
-                continue
+        async for thread in channel.archived_threads(limit=None):
+            await automod_thread(self, channel_id, thread, duration, results)
```
brtran4, line 29: i think we can combine the archived_threads and threads together and just loop once

src/workers/helper.py
```diff
@@ -1,11 +1,27 @@
...
 
+
+class UnableToDM(Exception):
+    pass
+
+
+@dataclass
+class AutomodResults:
+    num_removed: int = 0
+    num_errors: int = 0
```
brtran4, line 20: we could return num_removed and num_errors as a tuple in `automod_thread`

brtran4 (changes requested):

src/workers/forum_automod.py
OTCompa: Archived threads are classified separately from normal threads, so both need to be looped over. The two for loops also loop over different things. I had them separate since `channel.archived_threads()` is an async method that yields a generator while `channel.threads` is a list.  To my knowledge, I believe it'd be more effort than it's worth to combine the two if it is even possible.

src/workers/helper.py
OTCompa: Personally, I don't really think it's cleaner as a tuple. Here's a sample of what I had to change:
### `src/workers/forum_automod.py`
```python
    for channel_id, duration...:
        num_removed = 0
        num_errors = 0

        ...

        async for thread in...:
            num_removed, num_errors = await automod_thread(
                self, channel_id, thread, duration, num_removed, num_errors
            )
```

### `src/workers/helper.py`
```python
async def automod_thread(
    self,
    channel_id,
    thread: discord.Thread,
    duration: int,
    num_removed: int,
    num_errors: int,
):
    if thread.flags.pinned:
        return num_removed, num_errors

    ...

    if starter_message is not...:
        return num_removed, num_errors

    ....
    try:
        await thread.delete()
        ret = num_removed + 1, num_errors
        # code for dming user here
        return ret
    except...:
        return num_removed, num_errors + 1
```
If you think it's still better I'll add it though!

src/workers/forum_automod.py
brtran4: it looks like they have the same content though. we're passing the threads to `automod_thread` the exact same way and the function seems like it doesnt distinguish between the two different kinds of threads. you could do `threads = channel.archived_threads + channel.threads` and it should automatically combine the two lists

src/workers/helper.py
brtran4: imo i think it's fine because the benefit of having the class isn't really that big, but maybe @KingPieson has a differing opinion from me. while we're here we might as well type the return type in `automod_thread` function signature

src/workers/forum_automod.py
OTCompa: channel.archived_threads() does not give a list so I can't do that
```
TypeError: unsupported operand type(s) for +: 'async_generator' and 'list'
```

src/workers/helper.py
OTCompa: I'll make it a tuple then :nod:

src/workers/forum_automod.py
brtran4: Ew okay I might poke at it on my own but we can put this in for now

src/workers/helper.py
```diff
@@ -1,11 +1,20 @@
...
 import discord
+from discord.utils import snowflake_time
 from discord.ext.commands import Bot
 from settings import get_settings
 from util import create_interaction_embed_context
 from typing import Optional
 
+
+class UnableToDM(Exception):
+    pass
```
brtran4, line 13: i think this would go better in util.py and we can raise a `RunTime` error here

OTCompa: If we aren't using this and using `RunTime` error instead then shouldn't I just remove this completely instead of moving it to `util.py`? or am I misunderstanding

brtran4: Oh like move this to util.py then you can raise the RunTime error inside the class so it would raise a UnableToDm error and a RunTimeError

src/util.py
```diff
@@ -18,6 +18,10 @@ def __init__(self, name, value):
         self.value = value
 
 
+class UnableToDM(RuntimeError):
+    pass
```
brtran4, line 22: the way we have it now, i dont think this will still raise any type of error, because it raises the `UnableToDM` error, but it doesnt actually raise anything useful for us. A caller could go `UnableToDM("User does not have DMs turned on")`, etc. generally, if we want to raise an error, we should be putting an error message to make it clear, otherwise we don't need to make this class to raise an error

brtran4 (changes requested):

src/util.py
OTCompa: I don't really understand what the issue is. Since it inherits `RuntimeError` (previously `Exception`), it has all the attributes/methods of a `RuntimeError`. I needed this class specifically because I wanted to handle a failed DM case but still properly log any other unrelated errors as an error.

brtran4: per convo on discord an NB comment would be good

brtran4 (approved):

PR #147, opened 2024-11-04
Max character ban fail fix old

Alex-Gideon:
Handles the 512 reason character error and standardizes the messages and embeds.

- Embeds happen when a ban succeeds
- Notes whether or not a DM succeeds or fails
- Ephemeral messages are always sent, including ban status and dm status, and match the error/result given in an embed if it exists.
- Correctly logs output for backend debugging.

### DIFF
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -21,12 +21,13 @@ def create_ban_commands(bot: Bot) -> None:
         reason: str,
     ):
         """Ban the specified user."""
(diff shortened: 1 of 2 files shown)


Alex-Gideon:
Possible outputs:
Ban Fail DM Succeed
image
Ban Fail Dm Fail 
image
Ban Succeed DM Fail
image
Ban Succeed DM Succeed  (ignore the reason...)
image

PR #148, opened 2024-11-05
Max character ban fail fix

Alex-Gideon:
Ticket: MOD-15

Handles the 512 reason character error and standardizes the messages and embeds.

- Embeds happen when a ban succeeds
- Notes whether or not a DM succeeds or fails
- Ephemeral messages are always sent, including ban status and dm status, and match the error/result given in an embed if it exists.
- Correctly logs output for backend debugging.

Possible outputs:
Ban Fail DM Succeed
image
Ban Fail Dm Fail 
image
Ban Succeed DM Fail
image
Ban Succeed DM Succeed  (ignore the reason...)
image

Ticket: MOD-15


Alex-Gideon:
Ticket: MOD-15

clairefleuret (changes requested):
image
conflicts need addressed

clairefleuret (approved):
has my approval from end user perspective, @brtran4 or @OTCompa please review from coding side

.github/workflows/commit-check.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          TICKET_PATTERN="^\s*(Ticket:\s*)?([A-Z]+-[0-9]+|NA)\s*$"
+          TICKET_PATTERN="Ticket:\s*([A-Z]+-[0-9]+|NA)"
```
brtran4, line 16: when @OTCompa gets this change in, we can delete this change

src/commands/ban_commands.py
```diff
@@ -21,10 +21,19 @@ async def ban(
         reason: str,
     ):
         """Ban the specified user."""
-        async with create_response_context(interaction) as response_message:
-            async with create_logging_embed(
-                interaction, user=user, reason=reason
-            ) as logging_embed:
-                await ban_user(logging_embed, user, reason)
+        # result is a tuple of state and descriptio
```
brtran4, line 24:
```suggestion
        # result is a tuple of state and description
```

src/commands/ban_commands.py
```diff
@@ -21,10 +21,19 @@ async def ban(
...
+        result = await ban_user(user, reason)
```
brtran4, line 25: accessing result directly is hard to read in dense code like this, so let's initialize some vars before we go into the if statements

src/commands/ban_commands.py
```diff
@@ -21,10 +21,19 @@ async def ban(
...
 
-                response_message.set_string(f"Successfully banned {user.mention}")
+        if result[0]:  # ban succeeded dont create embed.
```
brtran4, line 27:
```suggestion
        if result[0]:  # ban succeeded
```
i think it's supposed to be like this? i'm kind of confused

src/commands/ban_commands.py
```diff
@@ -21,10 +21,19 @@ async def ban(
         reason: str,
     ):
         """Ban the specified user."""
-        async with create_response_context(interaction) as response_message:
```
brtran4, line 24: we still need the response context so we dont time out on the interaction

src/services/ban_service.py
```diff
@@ -1,20 +1,79 @@
...
-    log_info_and_embed(logging_embed, logger, f"going to ban {user.mention}")
+async def ban_user(
+    user: discord.Member, reason: str
+) -> Optional[Tuple[bool, bool, str]]:
+    """Executes ban of user
 
-    log_info_and_embed(logging_embed, logger, f"because {reason}")
+    Args:
+        user (discord.Member): mention or "@"/user ID of the user being banned.
+        reason (str): description of ban reason.
```
brtran4, line 18: let's mention that it has to be below 512 chars

src/services/ban_service.py
```diff
@@ -1,20 +1,79 @@
...
+            dm_state = True
+        except discord.Forbidden as e:
+            logger.error(f"DM to {user.mention} failed due to permission error: {e}")
+            dm_state = False
+        except discord.HTTPException as e:
+            logger.error(f"DM to {user.mention} due to HTTP Error: {e}")
+            dm_state = False
+        except Exception as e:
+            logger.error(f"DM to {user.mention} due to Unknown Error: {e}")
+            dm_state = False
```
brtran4, line 46: imo i feel like this amount of error handling is a bit extreme as it is cluttering up a lot of space than necessary. since we're already printing out the exception, and we're not particularly printing anything different/more useful than what the `Exception` object is doing, i think we can simplify to just use the catch-all `Exception`

src/services/ban_service.py
```diff
@@ -1,20 +1,79 @@
...
+            result = True
+        except discord.Forbidden as e:
+            logger.error(f"Ban of {user.mention} failed due to permission error: {e}")
+            result = False
+        except discord.HTTPException as e:
+            logger.error(f"Ban of {user.mention} failed due to HTTP Error: {e}")
+            result = False
+        except Exception as e:
+            logger.error(f"Ban of {user.mention} failed due to Unknown Error: {e}")
+            result = False
```
brtran4, line 61: same here

brtran4 (changes requested):

.github/workflows/commit-check.yaml
brtran4: the change is in, so we can get rid of this file now

src/services/ban_service.py
Alex-Gideon: Fixed in latest version

src/services/ban_service.py
Alex-Gideon: Fixed in latest version

Alex-Gideon:
@brtran4 I think i made the changes you asked for, mind re-reviewing at your earliest convenience?

src/commands/ban_commands.py
```diff
@@ -21,10 +21,21 @@ async def ban(
         reason: str,
     ):
         """Ban the specified user."""
+
```
brtran4, line 24: nit: let's get rid of this line for min-diff

src/commands/ban_commands.py
```diff
@@ -21,10 +21,21 @@ async def ban(
...
         async with create_response_context(interaction) as response_message:
-            async with create_logging_embed(
-                interaction, user=user, reason=reason
-            ) as logging_embed:
-                await ban_user(logging_embed, user, reason)
+            (ban_state, dm_state, result_description) = await ban_user(user, reason)
 
-                response_message.set_string(f"Successfully banned {user.mention}")
+            if ban_state:  # ban succeeded
+                if not dm_state:  # dm failed
```
brtran4, line 29: ~nit: with this naming convention, i think it reads better if we do `if dm_state is None`~. looking further down the code i see that this is a bool, bools should be typically named as `is_xxx` to make it more clear that it is a bool and not an object (for both `ban_state` and `dm_state`)

src/services/ban_service.py
```diff
@@ -1,20 +1,67 @@
...
+            - result (bool): Indicates whether the ban succeeded.
+            - dm_state (bool): Indicates whether the DM notification to the user succeeded.
+            - result_description (str): A message describing the status of the ban operation.
+    """
+    if len(reason) < 512:
+        result_description = ""
+        dm_state = False
+        # attempt to send a DM to the user with the reason for the ban
+        try:
+            await send_dm(
```
brtran4, line 31: we lost the TODO comment here

src/services/ban_service.py
```diff
@@ -1,20 +1,67 @@
...
+            result_description = f"Successfully banned {user.mention} and DM has been sent with ban reason."
+        else:  # full failure
+            result_description = f"Unable to ban {user.mention} and unable to send DM. Please ban via discord built-in ban feature or try again later."
+
+    else:
+        # reason too large, ban action canceled.
+        result_description = f"Unable to ban: {user.mention}, reason given is too long (above 512 characters). Please shorten ban reason."
+        result = False
+    # always return this.
+    return (result, dm_state, result_description)
```
brtran4, line 67: let's rename `result` to `ban_result`

brtran4 (changes requested):

Alex-Gideon:
Changes have been made @brtran4

src/commands/ban_commands.py
```diff
@@ -22,9 +22,19 @@ async def ban(
...
+                if not is_dm_sent:  # dm failed
+                    async with create_logging_embed(
+                        interaction, user=user, reason=reason, error=result_description
+                    ) as logging_embed:
+                        response_message.set_string(result_description)
+                else:  # ban succeeded, dm failed.
+                    async with create_logging_embed(
+                        interaction, user=user, reason=reason, result=result_description
+                    ) as logging_embed:
+                        response_message.set_string(result_description)
```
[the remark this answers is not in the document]
Alex-Gideon: result vs error as the name for result description changes the output of the embed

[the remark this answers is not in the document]
brtran4: ah i see. ok lgtm

brtran4 (approved):

PR #149, opened 2024-11-06
Update exile string formatting

KingPieson:
Update the formatting for exile duration to the form of (1-99)(day/min/sec/hour)

Ticket: MOD-13

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -31,7 +31,7 @@ def create_exile_commands(bot: Bot) -> None:
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(
         user="User being exiled",
-        duration="The duration of the exile. Duration should be in the form of [1 or 2 digits][s, d, m, h]",
+        duration="Enter a value between 1 and 99 followed by sec, min, hour, or day. Examples: 1sec, 1min, 1hour, 1day.",
         reason="Reason for exile",
     )
     async def exile(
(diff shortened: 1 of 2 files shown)


src/commands/exile_commands.py
```diff
@@ -31,7 +31,7 @@ async def unexile(interaction: discord.Interaction, user: discord.Member):
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(
         user="User being exiled",
-        duration="The duration of the exile. Duration should be in the form of [1 or 2 digits][s, d, m, h]",
+        duration="Enter a value between 1 and 99 followed by sec, min, hour, or day. Examples: 1sec, 1min, 1hour, 1day.",
```
brtran4, line 34: we should probably put a check to make sure that the value is between 1 and 99

brtran4 (changes requested):

src/commands/exile_commands.py
KingPieson: The regex requires the two digits to be at the start of the string, so you're not able to input a value like 500

brtran4: oh nice. i think we should still tell the user that if the regex cant find anything, then their input is wrong

brtran4 (changes requested):

src/commands/exile_commands.py
KingPieson: We have that set up, if the `calculate_time_delta` regex doesn't match the input, we return None and then respond with the ephemeral message below.

brtran4 (approved):

PR #152, opened 2024-11-08
Handle exceptions in automod

OTCompa:
Ticket: MOD-49
The automod task seems to have crashed yesterday because it didn't have permissions in the FRU channel. As a result, it did not reschedule the task and did not run today. I've added a block to handle any exception so the task is still scheduled if anything happens. I've also commented out the FRU channel for now since it won't be active till the ultimate releases


src/workers/forum_automod.py
```diff
@@ -27,43 +24,49 @@ async def autodelete_threads(self):
     for channel_id, duration in settings.automod_inactivity.items():
         num_removed = 0
         num_errors = 0
-        channel = guild.get_channel(channel_id)
-        if channel is None:
-            continue
+        try:
+            channel = guild.get_channel(channel_id)
+            if channel is None:
+                continue
```
KingPieson, line 30: might be good to log somewhere that we failed to get the channel

KingPieson (approved):

brtran4 (approved):

PR #154, opened 2024-11-08
Add black formatting to MakeFile

brtran4:
Ticket: MOD-50


OTCompa (approved):
it's just `py` on my system but my IDE automatically formats on save so it doesn't matter for me

PR #163, opened 2024-11-16
DM user when un-exiled

clairefleuret:
Ticket: MOD-77

User will now be DM'd when an exile expires

image


brtran4 (approved):
nit: this is linked to the wrong ticket

clairefleuret:
Ticket: MOD-80

PR #166, opened 2024-11-20
Add chunked message helper function and condense automod notification message

OTCompa:
Ticket: MOD-86
This PR condenses the automod notification message into a bulk ping. 

`send_chunked_message()` has been added to `util.py` that automatically splits messages that may exceed discord's character limit. `chunk_generator()` may also be used to extend functionality to embeds (3k characters max) if needed. I also fixed a small bug with error logging + made an error slightly more verbose

### DIFF
--- a/src/workers/helper.py
+++ b/src/workers/helper.py
@@ -45,7 +45,7 @@ async def automod_thread(
     duration: int,
     num_removed: int,
     num_errors: int,
-    notifying_channel: discord.TextChannel,
+    user_list: list[int],
 ):
     if thread.flags.pinned:
         # skip the for loop if the thread is pinned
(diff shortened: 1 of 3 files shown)


src/workers/helper.py
```diff
@@ -73,12 +73,8 @@ async def automod_thread(
...
         ret = num_removed + 1, num_errors
         if thread.owner is not None:
-            try:
-                await notifying_channel.send(
-                    f'{thread.owner.mention}, your post "{thread.name}" in <#{channel_id}> has been automatically deleted as {duration} days have passed without any activity or the original message has been deleted.'
-                )
-            except Exception as e:
-                raise UnableToNotify(e)
+            if thread.owner_id not in user_list:
+                user_list.append(thread.owner_id)
```
brtran4, line 77: would we need this check if we changed `user_list` into a `set`?

brtran4 (changes requested):
since this change is going into 1.0 hotfix 2, we'll need to target this to 1.0-hotfix. once this change goes in we'll have to `git merge` and push to `1.1` and `main`

brtran4 (approved):

PR #168, opened 2024-11-22
Clean up user model

KingPieson:
change user model to have validation on it via pydantic, to help with typing

Ticket: MOD-20

### DIFF
--- a/src/commands/strikes_command.py
+++ b/src/commands/strikes_command.py
@@ -15,4 +15,6 @@ def create_strikes_commands(bot: Bot) -> None:
         reason: str,
     ):
         """Add a strike to the user"""
-        await interaction.response.send_message("This command is not currently implemented", ephemeral=True)
+        await interaction.response.send_message(
+            "This command is not currently implemented", ephemeral=True
+        )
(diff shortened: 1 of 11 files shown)


src/commands/strikes_command.py
```diff
@@ -0,0 +1,20 @@
+import discord
+from discord.ext.commands import Bot
+from util import is_user_moderator
+from enums import StrikeSeverity
+
+
+def create_strikes_commands(bot: Bot) -> None:
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    @discord.app_commands.describe(user="User being exiled")
```
brtran4, line 10: `User being striked`? not sure about the wording here @clairefleuret

brtran4 (changes requested):

src/commands/strikes_command.py
clairefleuret: Why am I being mentioned

clairefleuret: Yeah being striked sounds fine

brtran4 (approved):

PR #170, opened 2024-11-22
Implement Slowmode Command

clairefleuret:
Ticket: MOD-53

- Adds the slash slowmode command
- Allows user to enter 0 (off) to 21600 (6hours) seconds for a slowmode interval
- Includes potential error handling and checks if channel is already set to the desired interval

### DIFF
--- a/src/bot.py
+++ b/src/bot.py
@@ -4,6 +4,7 @@ from settings import get_settings
 from commands.exile_commands import create_exile_commands
 from commands.ban_commands import create_ban_commands
 from commands.helper import create_bot_errors
+from commands.slowmode_commands import create_slowmode_commands
 import logging
 import workers
 
@@ -29,4 +30,5 @@ class ModdingwayBot(Bot):
         create_exile_commands(self)
         create_ban_commands(self)
         create_bot_errors(self)
+        create_slowmode_commands(self)
         logger.info("Registering commands finished")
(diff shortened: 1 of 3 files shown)


src/bot.py
```diff
@@ -29,4 +30,5 @@ def _register_commands(self):
         create_exile_commands(self)
         create_ban_commands(self)
         create_bot_errors(self)
+        create_slowmode_commands(self)
```
brtran4, line 33: nit: can we move this above `create_bot_errors`

src/commands/slowmode_commands.py
```diff
@@ -0,0 +1,31 @@
...
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+def create_slowmode_commands(bot: Bot) -> None:
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    @discord.app_commands.describe(
+        channel="Channel name slowmode is being modified in",
```
brtran4, line 17: maybe using `set` instead of `modified`

src/commands/slowmode_commands.py
```diff
@@ -0,0 +1,31 @@
...
+        interval="The number of seconds desired between each message. Must be between 0 (off) and 21600 (6 hours).",
+    )
+    async def set_slowmode(
+        interaction: discord.Interaction,
+        interval: int,
+        channel: discord.TextChannel = None,
```
brtran4, line 23: i think maybe we should have it the other way since it's more intuitive to set the channel and then the interval

PR #177, opened 2024-11-22
Implement slowmode command

clairefleuret:
Ticket: MOD-53

- Adds the slash slowmode command
- Allows user to enter 0 (off) to 21600 (6hours) seconds for a slowmode interval
- Includes potential error handling and checks if channel is already set to the desired interval

### DIFF
--- a/src/commands/slowmode_commands.py
+++ b/src/commands/slowmode_commands.py
@@ -1,0 +1,25 @@
+import discord
+import logging
+from discord.ext.commands import Bot
+from settings import get_settings
+from services.slowmode_service import edit_slowmode
+from util import is_user_moderator
+from .helper import create_logging_embed, create_response_context
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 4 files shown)


src/commands/slowmode_commands.py
```diff
@@ -0,0 +1,31 @@
...
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    @discord.app_commands.describe(
+        interval="The number of seconds desired between each message. Must be between 0 (off) and 21600 (6 hours).",
+        channel="Channel name slowmode is being modified in",
+    )
+    async def set_slowmode(
+        interaction: discord.Interaction,
+        interval: int,
+        channel: discord.TextChannel = None,
```
brtran4, line 23: er sorry i meant the other way where it should be channel before interval

brtran4 (changes requested):

brtran4 (approved):

PR #182, opened 2024-11-22
More robust role checks

clairefleuret:
Ticket: MOD-28

- Removes all unnecessary is_user_moderator roles and checks only if the user has "Mod" role.
- Better error handeling if the role is missing in is_user_moderator check
- `ERROR:commands.helper:An unexpected error has occurred The check functions for command` will no longer be reported if check is failed, gracefully catching this error

### DIFF
--- a/src/commands/helper.py
+++ b/src/commands/helper.py
@@ -30,20 +30,35 @@ def create_logging_embed(interaction: discord.Interaction, **kwargs):
         fields=fields,
     )
 
-
 def create_bot_errors(bot: Bot) -> None:
     @bot.tree.error
     async def on_app_command_error(interaction: discord.Interaction, error):
-        # Check if the error is due to a cooldown
+        # Handle CommandOnCooldown error
         if isinstance(error, discord.app_commands.CommandOnCooldown):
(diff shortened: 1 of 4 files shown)


src/commands/helper.py
```diff
@@ -34,15 +34,32 @@ def create_logging_embed(interaction: discord.Interaction, **kwargs):
 def create_bot_errors(bot: Bot) -> None:
     @bot.tree.error
     async def on_app_command_error(interaction: discord.Interaction, error):
-        # Check if the error is due to a cooldown
+        # Handle CommandOnCooldown error
         if isinstance(error, discord.app_commands.CommandOnCooldown):
             remaining_time = int(error.retry_after) + int(time.time())
             await interaction.response.send_message(
-                f"This command is on cooldown. Time remaining: <t:{remaining_time}:R>",
+                f"This command is on cooldown. Time remaining: <t:{remaining_time}:R>.",
```
brtran4, line 41: i think we can change this text to something like "Try again in" instead of "Time remaining"

brtran4 (approved):

PR #186, opened 2024-11-22
Fix ban command so staff cannot ban each other

clairefleuret:
Ticket: MOD-89

- Users cannot ban someone with equal or higher role their current role via moddingway.
- User is given the error when trying to ban someone of equal or higher rank: Unable to ban {user.mention}: You cannot ban a user with an equal or higher role than yourself
- Fixed README to be up to date with current scope of role requirements

### DIFF
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -15,46 +15,42 @@ def create_ban_commands(bot: Bot) -> None:
     @discord.app_commands.describe(
         user="User being banned",
         reason="Reason for ban",
-        delete_messages="Type 'Yes' to delete messages or 'No' to keep them (default: 'No')",
+        delete_messages="Whether messages from the banned user should be deleted or not",
     )
     async def ban(
         interaction: discord.Interaction,
         user: discord.Member,
         reason: str,
-        delete_messages: str = "No",  # Default to 'No' for no message deletion
+        delete_messages: bool = False,  # Default to false for no message deletion as we typically don't want to delete messages.
     ):
         """Ban the specified user."""
-        # Normalize and validate the input
-        delete_messages = delete_messages.strip().capitalize()  # Handle capitalization
-        if delete_messages not in ["Yes", "No"]:
-            await interaction.response.send_message(
-                "Invalid value for 'delete_messages'. Please enter 'Yes' to delete messages or 'No' to keep them.",
-                ephemeral=True,
-            )
-            return
-
-        # Convert delete_messages to a boolean for ban logic
-        delete_messages_flag = delete_messages == "Yes"
-
         async with create_response_context(interaction) as response_message:
-            (is_banned, result_description) = await ban_user(
-                interaction.user, user, reason, delete_messages_flag
-            )
-
-            if is_banned:  # Ban succeeded
-                async with create_logging_embed(
-                    interaction,
-                    user=user,
-                    reason=reason,
-                    result=result_description,
-                    delete_messages_flag=delete_messages_flag,  # Pass as "Yes" or "No"
-                ) as logging_embed:
-                    response_message.set_string(result_description)
-            else:  # Ban failed
-                async with create_logging_embed(
-                    interaction, user=user, reason=reason, error=result_description
-                ) as logging_embed:
-                    response_message.set_string(result_description)
+            ban_success, error = await ban_user(user, reason, delete_messages)
+            # Ensure invoking_member has a higher role position than the target user.
+            if user.top_role >= interaction.user.top_role:
+                response_message.set_string(
+                    f"Unable to ban {user.mention}: You cannot ban a user with an equal or higher role than yourself."
+                )
+                return
+            async with create_logging_embed(
+                interaction,
+                user=user,
+                reason=reason,
+                delete_messages=delete_messages,
+            ) as logging_embed:
+                if ban_success:
+                    success_str = f"Successfully banned {user.mention}."
+                    logging_embed.add_field(
+                        name="Result", value=success_str, inline=False
+                    )
+                    response_message.set_string(success_str)
+                    if error:
+                        logging_embed.add_field(
+                            name="DM Status", value=error, inline=False
+                        )
+                else:
+                    logging_embed.add_field(name="Error", value=error, inline=False)
+                    response_message.set_string(error)
 
     @bot.tree.context_menu(name="Ban User")
     @discord.app_commands.check(is_user_moderator)
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -37,7 +37,7 @@ def create_ban_commands(bot: Bot) -> None:
         delete_messages_flag = delete_messages == "Yes"
 
         async with create_response_context(interaction) as response_message:
-            (is_banned, is_dm_sent, result_description) = await ban_user(
+            (is_banned, result_description) = await ban_user(
                 interaction.user, user, reason, delete_messages_flag
             )
 
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -41,13 +41,12 @@ def create_ban_commands(bot: Bot) -> None:
 
             if is_banned:  # Ban succeeded
                 async with create_logging_embed(
-                    interaction, user=user, reason=reason, result=result_description
+                    interaction,
+                    user=user,
+                    reason=reason,
+                    result=result_description,
+                    delete_messages_flag=delete_messages_flag,
                 ) as logging_embed:
-                    logging_embed.add_field(
-                        name="Messages Deleted",
-                        value="Yes" if delete_messages_flag else "No",
-                        inline=False,
-                    )
                     response_message.set_string(result_description)
             else:  # Ban failed
                 async with create_logging_embed(
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -15,25 +15,27 @@ def create_ban_commands(bot: Bot) -> None:
     @discord.app_commands.describe(
         user="User being banned",
         reason="Reason for ban",
-        delete_messages="Type 'Y' to delete messages or 'N' to keep them (default: 'N')",
+        delete_messages="Type 'Yes' to delete messages or 'No' to keep them (default: 'No')",
     )
     async def ban(
         interaction: discord.Interaction,
         user: discord.Member,
         reason: str,
-        delete_messages: str = "N",  # Default to 'N' for no message deletion
+        delete_messages: str = "No",  # Default to 'No' for no message deletion
     ):
         """Ban the specified user."""
         # Normalize and validate the input
-        delete_messages = delete_messages.strip().upper()
-        if delete_messages not in ["Y", "N"]:
+        delete_messages = delete_messages.strip().capitalize()  # Handle capitalization
+        if delete_messages not in ["Yes", "No"]:
             await interaction.response.send_message(
-                "Invalid value for 'delete_messages'. Please enter 'Y' to delete messages or 'N' to keep them.",
+                "Invalid value for 'delete_messages'. Please enter 'Yes' to delete messages or 'No' to keep them.",
                 ephemeral=True,
             )
             return
 
-        delete_messages_flag = delete_messages == "Y"  # Convert to boolean
+        # Convert delete_messages to a boolean for ban logic
+        delete_messages_flag = delete_messages == "Yes"
+
         async with create_response_context(interaction) as response_message:
             (is_banned, is_dm_sent, result_description) = await ban_user(
                 interaction.user, user, reason, delete_messages_flag
@@ -45,7 +47,7 @@ def create_ban_commands(bot: Bot) -> None:
                     user=user,
                     reason=reason,
                     result=result_description,
-                    delete_messages_flag=delete_messages_flag,
+                    delete_messages_flag=delete_messages_flag,  # Pass as "Yes" or "No"
                 ) as logging_embed:
                     response_message.set_string(result_description)
             else:  # Ban failed
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -15,16 +15,18 @@ def create_ban_commands(bot: Bot) -> None:
     @discord.app_commands.describe(
         user="User being banned",
         reason="Reason for ban",
+        delete_messages="Whether to delete the user's messages (default: False)",
     )
     async def ban(
         interaction: discord.Interaction,
         user: discord.Member,
         reason: str,
+        delete_messages: bool = False,  # Default to not deleting messages
     ):
         """Ban the specified user."""
         async with create_response_context(interaction) as response_message:
             (is_banned, is_dm_sent, result_description) = await ban_user(
-                interaction.user, user, reason
+                interaction.user, user, reason, delete_messages
             )
 
             if is_banned:  # ban succeeded
--- a/src/commands/ban_commands.py
+++ b/src/commands/ban_commands.py
@@ -23,7 +23,9 @@ def create_ban_commands(bot: Bot) -> None:
     ):
         """Ban the specified user."""
         async with create_response_context(interaction) as response_message:
-            (is_banned, is_dm_sent, result_description) = await ban_user(user, reason)
+            (is_banned, is_dm_sent, result_description) = await ban_user(
+                interaction.user, user, reason
+            )
 
             if is_banned:  # ban succeeded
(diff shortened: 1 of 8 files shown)


src/services/ban_service.py
```diff
@@ -9,60 +9,55 @@
...
-            result_description = (
-                f"Successfully banned {user.mention} but DM failed to send."
-            )
-        elif ban_result and dm_state:  # full success
-            result_description = f"Successfully banned {user.mention} and DM has been sent with ban reason."
-        else:  # full failure
-            result_description = f"Unable to ban {user.mention} and unable to send DM. Please ban via discord built-in ban feature or try again later."
-
-    else:
-        # reason too large, ban action canceled.
```
brtran4, line 64: i dont see actual changes in here, wonder why this is being replaced by itself

brtran4 (changes requested):
we'll also need to edit the ban user modal in `/src/ui/ban_modal.py`

clairefleuret:
Fixing so messages are not auto deleted. An optional field to remove messages to be added.

src/services/ban_service.py
clairefleuret: not sure tbh

src/commands/ban_commands.py
```diff
@@ -15,15 +15,19 @@ def create_ban_commands(bot: Bot) -> None:
     @discord.app_commands.describe(
         user="User being banned",
         reason="Reason for ban",
+        delete_messages="Whether to delete the user's messages (default: False)",
     )
     async def ban(
         interaction: discord.Interaction,
         user: discord.Member,
         reason: str,
+        delete_messages: bool = False,  # Default to not deleting messages
```
brtran4, line 24: Doesn't the documentation say that to delete messages you need to say how many seconds worth of messages to delete?

src/services/ban_service.py
```diff
@@ -9,60 +9,61 @@
...
-        result_description = f"Unable to ban: {user.mention}, reason given is too long (above 512 characters). Please shorten ban reason."
-        ban_result = False
-    # always return this.
-    return (ban_result, dm_state, result_description)
+    try:
+        # Delete messages only if delete_messages is True
+        delete_days = 7 if delete_messages else 0
+        await user.ban(reason=reason, delete_message_days=delete_days)
```
brtran4, line 60: We should use seconds instead of days since it's deprecated and will be removed at some point

src/commands/ban_commands.py
clairefleuret: its hardcoded in to only be the same value. we will not give the option since there is really no need.

the only reason we would want to remove messages is for bots that comin. a week will be more than enough to delete spam messages

src/services/ban_service.py
clairefleuret: done

brtran4:
hrm i feel like we should uniformly decide if we want `delete_messages` to be either true/false or y/n, not both

src/services/ban_service.py
```diff
@@ -9,60 +9,62 @@
...
 
-    else:
-        # reason too large, ban action canceled.
-        result_description = f"Unable to ban: {user.mention}, reason given is too long (above 512 characters). Please shorten ban reason."
-        ban_result = False
-    # always return this.
-    return (ban_result, dm_state, result_description)
+    try:
+        # Delete messages only if delete_messages is True
+        # 604800 seconds is the maximum value for delete_message_seconds, and is equivalent to 7 days.
```
brtran4, line 59: can we describe a little more in detail why we're doing this in here (just for people who read this in the future)

clairefleuret:
image
image
image
image
image
image

src/services/ban_service.py
clairefleuret: done

clairefleuret:
> Fixing so messages are not auto deleted. An optional field to remove messages to be added.

done

clairefleuret:
> hrm i feel like we should uniformly decide if we want `delete_messages` to be either true/false or y/n, not both

done

src/commands/ban_commands.py
```diff
@@ -15,30 +15,45 @@ def create_ban_commands(bot: Bot) -> None:
...
-        response_message.set_string(result_description)
+            if is_banned:  # Ban succeeded
+                async with create_logging_embed(
+                    interaction, user=user, reason=reason, result=result_description
+                ) as logging_embed:
+                    logging_embed.add_field(
+                        name="Messages Deleted",
+                        value="Yes" if delete_messages_flag else "No",
+                        inline=False,
+                    )
```
brtran4, line 50: since this is present in the / command and modal, let's move this inside `ban_service.py` and keep the `response_message` object inside

src/ui/ban_modal.py
```diff
@@ -18,14 +18,41 @@ def __init__(self, user: discord.Member) -> None:
...
+                )
 
-                response_message.set_string(f"Successfully banned {self.user.mention}")
+                if result[0]:  # Ban was successful
+                    response_message.set_string(
+                        f"Successfully banned {self.user.mention}."
+                    )
+                else:  # Ban failed
+                    response_message.set_string(result[2])
+                    logging_embed.add_field(name="Error", value=result[2], inline=False)
```
brtran4, line 58: in general we should not use result directly. we should assign `result[0]` and `result[2]` to their own variables

brtran4 (changes requested):

src/commands/ban_commands.py
```diff
@@ -15,30 +15,45 @@ def create_ban_commands(bot: Bot) -> None:
...
+        # Normalize and validate the input
+        delete_messages = delete_messages.strip().upper()
+        if delete_messages not in ["Y", "N"]:
+            await interaction.response.send_message(
+                "Invalid value for 'delete_messages'. Please enter 'Y' to delete messages or 'N' to keep them.",
+                ephemeral=True,
+            )
+            return
+
+        delete_messages_flag = delete_messages == "Y"  # Convert to boolean
```
brtran4, line 36: what about the case where a user inputs `y`?

clairefleuret: its not case sensitive

src/commands/ban_commands.py
clairefleuret: done

src/commands/helper.py
```diff
@@ -22,6 +24,15 @@ def create_logging_embed(interaction: discord.Interaction, **kwargs):
...
 
+    # Add "Messages Deleted" field if delete_messages_flag is provided
+    if delete_messages_flag is not None:
+        fields.append(
+            EmbedField(
+                "Messages Deleted",
+                "Yes" if delete_messages_flag else "No",
+            )
+        )
+
```
brtran4, line 35: i think this should be included in `kwargs`. `kwargs` is basically a generic catch all variable to deal with all the various cases such as the match/case statements above. we should also throw this up there as well using `kwargs`

brtran4 (changes requested):

src/commands/helper.py
clairefleuret: i cant really add a way to do this so the output will be a little eh.

We run into a pylance issue since there would be irrefutable cases and python only allows 1 in match/case.

image

brtran4: you could make an object that holds some more information just like in the discord.Member case

brtran4: Or what you could do i sjust replace the title with actual text instead of using the key

clairefleuret: how would i replace the title with actual text

brtran4:
``` 
                case _:  # Irrefutable pattern for all other types
                    if "delete_messages_flag" in key:
                        fields.append(EmbedField("Messages Deleted", value))
                    else:
                        fields.append(EmbedField(key.title(), value))
```

src/commands/ban_commands.py
```diff
@@ -15,30 +15,46 @@ def create_ban_commands(bot: Bot) -> None:
...
-                        response_message.set_string(result_description)
-                else:  # ban succeeded, dm failed.
-                    async with create_logging_embed(
-                        interaction, user=user, reason=reason, result=result_description
-                    ) as logging_embed:
-                        response_message.set_string(result_description)
-            else:  # ban failed, dont create embed
-                response_message.set_string(result_description)
-        response_message.set_string(result_description)
+            (is_banned, is_dm_sent, result_description) = await ban_user(
```
brtran4, line 40: since we moved `is_dm_sent` into the ban, we don't really need this out here (aka it's not being used in this function anymore). let's change the function to only return `is_banned` and `result_description`

brtran4 (changes requested):

src/commands/ban_commands.py
clairefleuret: done

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
 logger = logging.getLogger(__name__)
 
 
 async def ban_user(
-    user: discord.Member, reason: str
+    invoking_member: discord.Member,
+    user: discord.Member,
+    reason: str,
+    delete_messages_flag: bool,
 ) -> Optional[Tuple[bool, bool, str]]:
```
brtran4, line 17: we'll need to edit this to just be `bool, str`

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
+        delete_messages_flag (bool): Whether to delete the user's messages.
 
     Returns:
-        Optional[Tuple[bool, bool, str]]: tuple containing (ban_result,dm_state,result_description)
-            - ban_result (bool): Indicates whether the ban succeeded.
-            - dm_state (bool): Indicates whether the DM notification to the user succeeded.
-            - result_description (str): A message describing the status of the ban operation.
+        Optional[Tuple[bool, bool, str]]: Result of the ban operation. Tuple contains:
+            - bool: True if ban was successful, False otherwise.
+            - bool: True if DM was successfully sent, False otherwise.
```
brtran4, line 29: remove

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
+        # 604800 seconds is the maximum value for delete_message_seconds, and is equivalent to 7 days.
+        delete_seconds = 604800 if delete_messages_flag else 0
+        await user.ban(reason=reason, delete_message_seconds=delete_seconds)
+        logger.info(f"Successfully banned {user.mention}")
+        return (True, dm_state, f"Successfully banned {user.mention}.")
+    except Exception as e:
+        logger.error(f"Failed to ban {user.mention}: {e}")
+        return (
+            False,
+            dm_state,
```
brtran4, line 78: remove this as well

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
+
+    try:
+        # We typically do NOT want to delete messages, as we want to preserve evidence.
+        # However, we may want to delete messages in cases where the user has posted inappropriate content or spam.
+        # Delete messages only if delete_messages is True
+        # 604800 seconds is the maximum value for delete_message_seconds, and is equivalent to 7 days.
+        delete_seconds = 604800 if delete_messages_flag else 0
+        await user.ban(reason=reason, delete_message_seconds=delete_seconds)
+        logger.info(f"Successfully banned {user.mention}")
+        return (True, dm_state, f"Successfully banned {user.mention}.")
```
brtran4, line 73: also edit to remove dm_state

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
+        await send_dm(
+            user,
+            f"Hello {user.display_name},\n\n"
+            "You are being informed that you have been **banned** from **NA Ultimate Raiding - FFXIV**.\n\n"
+            "**Reason for the ban:**\n"
+            f"> {reason}\n\n"
+            f"If you believe this ban was issued in error you can reach out to the Moderation Team. Otherwise, you may appeal this ban starting on <t:{appeal_deadline}:F>.\n\n"
+            "Please note that any further attempts to rejoin the server will be met with a permanent ban.\n\n",
+        )
+        dm_state = True
```
brtran4, line 61: we don't need this anymore

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
-            logger.info(f"Successfully banned {user.mention}")
-            ban_result = True
-        except Exception as e:
-            logger.error(f"Ban of {user.mention} failed due to error: {e}")
-            ban_result = False
+    # Ensure invoking_member has a higher role position than the target user.
+    if user.top_role >= invoking_member.top_role:
+        return (
+            False,
+            False,
```
brtran4, line 43: remove

src/services/ban_service.py
```diff
@@ -3,66 +3,78 @@
...
-            )
-            logger.info(f"Successfully sent dm to {user.mention}")
-            dm_state = True
-        except Exception as e:
-            logger.error(f"DM to {user.mention} due failed due to error: {e}")
-            dm_state = False
+    if len(reason) >= 512:
+        return (
+            False,
+            False,
```
brtran4, line 35: remove

brtran4 (changes requested):

src/services/ban_service.py
```diff
@@ -3,66 +3,73 @@
...
+        user (discord.Member): The user being banned.
+        reason (str): Reason for the ban.
+        delete_messages_flag (bool): Whether to delete the user's messages.
 
     Returns:
-        Optional[Tuple[bool, bool, str]]: tuple containing (ban_result,dm_state,result_description)
-            - ban_result (bool): Indicates whether the ban succeeded.
-            - dm_state (bool): Indicates whether the DM notification to the user succeeded.
-            - result_description (str): A message describing the status of the ban operation.
+        Optional[Tuple[bool, bool, str]]: Result of the ban operation. Tuple contains:
```
brtran4, line 27: take out the extra bool

src/services/ban_service.py
```diff
@@ -3,66 +3,73 @@
...
+            - bool: True if ban was successful, False otherwise.
+            - str: Description of the result of the ban operation
```
brtran4, line 29: we can leave the variable names in here

src/services/ban_service.py
```diff
@@ -3,66 +3,73 @@
...
+    # Calculate the timestamp for 30 days from now
+    appeal_deadline = int((datetime.now(timezone.utc) + timedelta(days=30)).timestamp())
 
-    else:
-        # reason too large, ban action canceled.
-        result_description = f"Unable to ban: {user.mention}, reason given is too long (above 512 characters). Please shorten ban reason."
-        ban_result = False
-    # always return this.
-    return (ban_result, dm_state, result_description)
+    dm_state = False
```
brtran4, line 47: remove

brtran4 (changes requested):

brtran4 (dismissed):

src/services/ban_service.py
```diff
@@ -3,66 +3,62 @@
...
-) -> Optional[Tuple[bool, bool, str]]:
+    user: discord.Member, reason: str, delete_messages: bool
+) -> Optional[Tuple[bool, str]]:
     """Executes ban of user.
 
-    Args:
-        user (discord.Member): mention or "@"/user ID of the user being banned.
-        reason (str): description of ban reason. Length of reason must be less than 512 characters
+     Args:
+         invoking_member (discord.Member): The moderator initiating the ban.
```
brtran4, line 18: nit: we should get rid of this description if we're not passing this in anymore

src/services/ban_service.py
```diff
@@ -3,66 +3,62 @@
...
-            result_description = f"Unable to ban {user.mention}, please ban via discord built-in ban feature. A DM has been sent to {user.mention} with ban reason."
-        elif ban_result and not dm_state:  # ban succeed dm fail.
-            result_description = (
-                f"Successfully banned {user.mention} but DM failed to send."
-            )
-        elif ban_result and dm_state:  # full success
-            result_description = f"Successfully banned {user.mention} and DM has been sent with ban reason."
-        else:  # full failure
-            result_description = f"Unable to ban {user.mention} and unable to send DM. Please ban via discord built-in ban feature or try again later."
+    dm_failed = None
```
brtran4, line 36: let's rename this to `dm_failed_message` just for clarity

brtran4 (changes requested):

src/services/ban_service.py
clairefleuret: done

src/services/ban_service.py
clairefleuret: done

brtran4 (approved):

PR #187, opened 2024-11-24
Add strike with automod

KingPieson:
Ticket: MOD-20


src/services/exile_service.py
```diff
@@ -76,7 +74,9 @@ async def exile_user(
             logging_embed, logger, f"Failed to send DM to exiled user, {e}"
         )
 
-    log_info_and_embed(logging_embed, logger, f"<@{user.id}> was successfully exiled")
+    log_info_and_embed(
+        logging_embed, logger, f"<@{user.id}> was successfully exiled for {duration}"
```
brtran4, line 78: while we're here i think we can also include an end time with the unix timestamp like on line 70

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
             logger,
-            f"User not found in database, creating new record",
+            "User not found in database, creating new record",
         )
-        db_user_id = users_database.add_user(user.id)
-        logger.info(f"Created user record in DB with id {db_user_id}")
-
-        db_user = create_empty_user(db_user_id, user.id, None, None)
+        db_user = users_database.add_user(user.id)
+        logger.info(f"Created user record in DB with id {db_user.user_id}")
```
brtran4, line 30: since we're logging this so many times throughout the code base, should we just throw this in `add_user`?

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
     # increment user points, update
+    db_user.last_infraction_timestamp = strike_timestamp
+    previous_points = db_user.get_strike_points()
+    _apply_strike_point_penalty(db_user, severity)
+    users_database.update_user_strike_points(db_user)
+
+    log_info_and_embed(
+        logging_embed,
+        logger,
+        f"<@{user.id}> was given a strike with ID {strike.strike_id}, bringing them to {db_user.get_strike_points()} points",
```
brtran4, line 55: nit: since the ID is in the footer, we don't need the ID here anymore

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
+        await exile_service.exile_user(
+            logging_embed, user, timedelta(days=3), exile_reason
+        )
+    elif total_points >= 3 and previous_points < 3:
+        punishment = "1 day exile"
+        await exile_service.exile_user(
+            logging_embed, user, timedelta(days=1), exile_reason
+        )
+    else:
+        punishment = "nothing"
```
brtran4, line 139: nit: let's match the style of how we're logging everything else and uppercase the n to be `Nothing`

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
+    )
+
+    punishment = await _apply_punishment(logging_embed, user, db_user, previous_points)
+    logging_embed.add_field(name="Punishment", value=punishment)
+
+    # message user
+    try:
+        await send_dm(
+            user,
+            f"Your actions in NA Ultimate Raiding - FFXIV resulted in a strike against your account. This may result in punishment depending on frequency or severity of your strikes.\n**Reason:** {reason}",
```
brtran4, line 65: nit:
```suggestion
            f"Your actions in NA Ultimate Raiding - FFXIV resulted in a strike against your account. This may result in punishment depending on the frequency or severity of your strikes.\n**Reason:** {reason}",
```

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
+        )
+    except Exception as e:
+        log_info_and_embed(
+            logging_embed, logger, f"Failed to send DM to exiled user, {e}"
+        )
+
+
+MINOR_INFRACTION_POINTS = 1
+MODERATE_INFRACTION_POINTS = 3
+SERIOUS_INFRACTION_POINTS = 7
```
brtran4, line 75: would this warrant a move to `constants.py`?

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
+
+    # TODO: known error, if an exiled user is given a strike, the follow up exile is not created
+    exile_reason = (
+        "Your actions were severe or frequent enough for you to receive this exile"
+    )
+
+    if total_points >= 15:
+        punishment = "permanent ban"
+        await ban_service.ban_user(
+            user, "Your strike were severe or frequent to be removed from the server"
```
brtran4, line 110: nit: replace `the server` with `NA Ultimate Raiding - FFXIV`

src/services/strike_service.py
```diff
@@ -13,22 +16,126 @@ async def add_strike(
...
+) -> str:
+    total_points = db_user.get_strike_points()
+
+    # TODO: known error, if an exiled user is given a strike, the follow up exile is not created
+    exile_reason = (
+        "Your actions were severe or frequent enough for you to receive this exile"
+    )
+
+    if total_points >= 15:
+        punishment = "permanent ban"
```
brtran4, line 108: nit: `Permanent ban`

brtran4 (changes requested):

src/services/strike_service.py
KingPieson: I'm unsure which ID you're talking about, I pulled out the strike ID tho

brtran4: yeah thats the one i was talking about mb

src/constants.py
```diff
@@ -0,0 +1,4 @@
+# Strikes constants
+MINOR_INFRACTION_POINTS = 1
+MODERATE_INFRACTION_POINTS = 3
+SERIOUS_INFRACTION_POINTS = 7
```
brtran4, line 4: hrm... i think since we already have an `enums.py` i think we should rename it to the new `constants.py` and move these three consts into that. can you do that in another PR/make a new ticket for that if you dont feel like doing it now

src/services/strike_service.py
```diff
@@ -13,22 +21,121 @@ async def add_strike(
...
+
+async def _apply_punishment(
+    logging_embed: discord.Embed,
+    user: discord.Member,
+    db_user: User,
+    previous_points: int,
+) -> str:
+    total_points = db_user.get_strike_points()
+
+    # TODO: known error, if an exiled user is given a strike, the follow up exile is not created
```
brtran4, line 101: if we're not going to handle this now, let's attach a ticket to this

KingPieson: Ticket added MOD-93 also called out the ticket number in this todo

src/constants.py
KingPieson: ticket created, MOD-94

PR #188, opened 2024-11-24
Take out excessive user in db check for unexile

oytundd:
Ticket: MOD-81
Bot returns an error if target trying to get unexiled does not exist in database.
Note: I am not sure if there is a condition where user exists in users table but not in the exiles so I could not get the next if statement to trigger. Please review.

### DIFF
--- a/src/services/exile_service.py
+++ b/src/services/exile_service.py
@@ -108,15 +108,13 @@ async def unexile_user(
     # update exile record
     db_user = users_database.get_user(user.id)
     if db_user is None:
+        error_message = "User does not exist in database, does not have any exiles, no action will be taken"
         log_info_and_embed(
             logging_embed,
             logger,
-            f"User not found in database, creating new record",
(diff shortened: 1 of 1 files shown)


KingPieson (approved):

KingPieson:
You should point this to the hotfix branch, I tried to do it myself and got some conflicts so I'm going to leave you to it

src/services/exile_service.py
```diff
@@ -108,15 +108,13 @@ async def unexile_user(
     # update exile record
     db_user = users_database.get_user(user.id)
     if db_user is None:
+        error_message = "User does not exist in database, does not have any exiles, no action will be taken"
```
brtran4, line 111: i think we can just shorten this string to `User does not have any exiles, no action will be taken`. if the user doesn't exist in the db that means they don't have an exile to begin with

brtran4 (approved):

PR #190, opened 2024-11-25
Decommission modals

clairefleuret:
Ticket: MOD-92

Decommissions modals that are no longer needed.


brtran4 (approved):
i can't tell if the ui folder got deleted too on github, but let's make sure to do that too

PR #191, opened 2024-11-25
Quick fix

OTCompa:
Ticked: NA
I asked to remove this earlier when I was tired, did a lookover after getting some sleep and we def want that


src/commands/helper.py
```diff
@@ -22,8 +22,8 @@ def create_logging_embed(interaction: discord.Interaction, **kwargs):
                     fields.append(EmbedField(key.title(), f"<@{value.id}>"))
                 case discord.ChannelType:
                     fields.append(EmbedField(key.title(), f"<#{value}>"))
-                case builtins.bool:
```
brtran4, line 25: if we're removing this let's also get rid of the import

brtran4 (changes requested):

brtran4 (approved):

PR #192, opened 2024-11-26
Adjust exile reporting

Alex-Gideon:
Ticket: MOD-41
 
 modify such that its: view instead of check

/view_active_exiles

prints users that are currently exiled

/view_exiles {user}

all historic non-roulette exiles.

needs another PR afterward to modify the embeds.


src/database/models/exile.py
```diff
@@ -1,9 +1,17 @@
...
+        user_id,
+        discord_id,
+        reason,
+        exile_status,
+        start_timestamp,
+        end_timestamp,
     ):
         self.exile_id = exile_id
         self.user_id = user_id
+        self.discord_id = (discord_id,)
```
KingPieson, line 14: I don't think you want this to be a tuple, I think it should just be `discord_id`

src/services/exile_service.py
```diff
@@ -142,18 +143,59 @@ async def get_user_exiles(logging_embed: discord.Embed, user: discord.User) -> s
...
-        exile_start_date = exile[2].strftime(TIME_FORMAT)
-        exile_end_date = exile[3].strftime(TIME_FORMAT) if exile[3] else "Indefinite"
-        exile_type = ExileStatus(exile[4]).name
+        logger.info(type(exile))
+        exile_id = exile.exile_id
+        if len(exile.discord_id) == 1:
+            discord_id = f"<@{exile.discord_id[0]}>"
+        else:
+            logger.error("Multiple or non 1 discord ID matched to userID.")
+            discord_id = exile.discord_id
```
KingPieson, line 186: I think you don't need this if you make the discord_id on the model not a tuple. I don't think there's a reason why there should be more than one discord user ID for an exile, at least with the assumption that we only have 1 server

src/services/exile_service.py
```diff
@@ -142,18 +143,59 @@ async def get_user_exiles(logging_embed: discord.Embed, user: discord.User) -> s
...
+    return result
+
 
-    result = f"Exiles found for <@{user.id}>:"
+async def get_active_exiles() -> str:
+    logger.info("Starting to run database command")
+    exile_list = exiles_database.get_all_active_exiles()
+    logger.info("db command done")
+    logger.info(type(exile_list))
+    logger.info(len(exile_list))
```
KingPieson, line 174: can we clean up these logs a little, we probably only need one here, if that

src/database/models/exile.py
```diff
@@ -1,9 +1,17 @@
 class Exile(object):
```
KingPieson, line 1: nit, and you can probably ignore this, but I'd like to move our DB models away from extending `object` and instead extend pydantic's `BaseModel` like the User is set up. I don't think you should do this here, just felt like commenting because you're updating here

src/database/exiles_database.py
```diff
@@ -107,27 +134,29 @@ def get_user_exiles(user_id) -> List[tuple]:
...
 
     with conn.get_cursor() as cursor:
         query = """
-        SELECT e.exileID, u.userID, u.discordUserID, e.endTimestamp
+        SELECT e.exileID, u.userID, u.discordUserID, e.reason, e.exileStatus, e.startTimestamp, e.endTimestamp
         FROM exiles e
         JOIN users u ON e.userID = u.userID
-        WHERE u.userID = %s AND  e.exileStatus = %s 
-        LIMIT 1;
+        WHERE e.exileStatus = %s AND e.reason != 'roulette';
```
KingPieson, line 151: Are you planning on following up here to change this from the reason to the createdBy that we talked about? I don't think that change is a blocker just curious

src/commands/exile_commands.py
```diff
@@ -105,11 +109,21 @@ async def roulette(interaction: discord.Interaction):
...
         async with create_response_context(interaction) as response_message:
             async with create_logging_embed(interaction, user=user) as logging_embed:
                 msg = await get_user_exiles(logging_embed, user)
 
                 response_message.set_string(msg)
+
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
+    async def view_active_exiles(interaction: discord.Interaction):
+        """Check all active exiles."""
```
KingPieson, line 124: nit can this explicitly call out that it's for all users that are exiled? Something like "Check all active exiles for all users."

src/database/exiles_database.py
```diff
@@ -90,12 +93,36 @@ def get_pending_unexiles() -> list[PendingExile]:
...
+        res = cursor.fetchone()
+
+        if res is not None:
+            return PendingExile(*res)
+        else:
+            return None
+
+
+#   exile_id, user_id, discord_id, reason, exile_status, start_timestamp, end_timestamp
+def get_user_exiles(user_id) -> Exile:
```
KingPieson, line 120: Can you update the return type here to `List[Exile]`

src/database/exiles_database.py
```diff
@@ -107,27 +134,29 @@ def get_user_exiles(user_id) -> List[tuple]:
         cursor.execute(query, params)
         res = cursor.fetchall()
 
-        return res
+        if res is not None:
+            return [Exile(*row) for row in res]  # Create a list of Exile objects
+        else:
+            return None
```
KingPieson, line 140: Can this return an empty list instead of null

KingPieson (changes requested):

src/database/models/exile.py
Alex-Gideon: image
output looked like this when it wasnt

Alex-Gideon: nevermind i see the error.

src/database/exiles_database.py
Alex-Gideon: not in this ticket, i think thats better suited to another one.. especially given the limited time I have before I leave + FRU

src/database/exiles_database.py
Alex-Gideon: Fixed in latest

src/database/exiles_database.py
Alex-Gideon: Fixed in latest

src/commands/exile_commands.py
Alex-Gideon: Fixed in latest

src/services/exile_service.py
Alex-Gideon: Fixed in latest

src/database/exiles_database.py
brtran4: if we're not doing it here then can you make a ticket for it?

KingPieson (approved):

src/database/exiles_database.py
Alex-Gideon: Created: https://naurffxiv.atlassian.net/browse/MOD-108

src/database/exiles_database.py
```diff
@@ -107,27 +134,29 @@ def get_user_exiles(user_id) -> List[tuple]:
...
+        SELECT e.exileID, u.userID, u.discordUserID, e.reason, e.exileStatus, e.startTimestamp, e.endTimestamp
         FROM exiles e
         JOIN users u ON e.userID = u.userID
-        WHERE u.userID = %s AND  e.exileStatus = %s 
-        LIMIT 1;
+        WHERE e.exileStatus = %s AND e.reason != 'roulette';
         """
 
-        params = (user_id, ExileStatus.TIMED_EXILED)
+        params = (ExileStatus.TIMED_EXILED,)
```
brtran4, line 154: i dont think we need the parentheses and the comma

src/services/exile_service.py
```diff
@@ -142,18 +143,52 @@ async def get_user_exiles(logging_embed: discord.Embed, user: discord.User) -> s
...
+        for exile in exile_list:
+            exile_id = exile.exile_id
+            exile_reason = exile.reason
+            exile_start_epoch = round(exile.start_timestamp.timestamp())
+            exile_end_epoch = (
+                round(exile.end_timestamp.timestamp()) if exile.exile_status else None
+            )
+            exile_start_date = f"<t:{exile_start_epoch}:F>"
+            exile_end_date = (
+                f"<t:{exile_end_epoch}:F>" if exile_end_epoch else "Indefinite"
```
brtran4, line 157: don't think we're supporting indefinite anymore

brtran4 (changes requested):

src/database/exiles_database.py
KingPieson: We do, params needs to be a tuple even if it's one item, and the parens plus comma is how to make a tuple with a single value

brtran4: ew that is disgusting lol

src/services/exile_service.py
KingPieson: We might still have some records in the DB that are indefinite, I'm not sure. We should eventually go through and just make those be 1 year and remove the indefinite corner cases IMO

brtran4: https://naurffxiv.atlassian.net/browse/MOD-109

brtran4 (approved):

src/database/exiles_database.py
KingPieson: what's worse is when you accidentally make a non-tuple into a tuple because of random parans or commas, then wonder why your code is fucked. Nova did that exact thing in this PR initially lol

PR #195, opened 2024-11-27
Update durations to be in Epoch Unix Timestamp and more

oytundd:
Ticket: MOD-82
The /exile and /roulette log embed now have expiration field showing dynamic timestamp.They also have nicer text in the duration field.

### DIFF
--- a/src/commands/exile_commands.py
+++ b/src/commands/exile_commands.py
@@ -7,7 +7,7 @@ from services.exile_service import (
     unexile_user,
     get_user_exiles,
     get_active_exiles,
-    nicer_duration,
+    format_time_string,
 )
 from util import is_user_moderator, calculate_time_delta
 from typing import Optional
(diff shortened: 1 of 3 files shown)


src/services/exile_service.py
```diff
@@ -193,3 +194,27 @@ async def get_active_exiles() -> str:
         )
 
     return result
+
+
+def nicer_duration(duration_string: str):
```
brtran4, line 199: let's rename this to `format_time_string`

src/commands/exile_commands.py
```diff
@@ -77,6 +85,9 @@ async def roulette(interaction: discord.Interaction):
         safety_choice = choice(safety_options)
         duration_choice = choice(exile_duration_options)
         duration_string = f"{duration_choice}hour"
```
brtran4, line 87: we should use the new function on this line

src/commands/exile_commands.py
```diff
@@ -53,10 +55,16 @@ async def exile(
                 ephemeral=True,
             )
             return
-
+        nicer_duration_str = nicer_duration(duration)
```
brtran4, line 58: don't think we need a new variable for this. we can just convert the duration on line 65

brtran4 (changes requested):
let's target this to 1.1-hotfix

brtran4 (approved):

PR #197, opened 2024-11-27
Update strikes_command.py

clairefleuret:
Ticket: MOD-110

image

### DIFF
--- a/src/commands/strikes_command.py
+++ b/src/commands/strikes_command.py
@@ -32,12 +32,12 @@ def create_strikes_commands(bot: Bot) -> None:
     @bot.tree.command()
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(user="User whose strikes you are checking")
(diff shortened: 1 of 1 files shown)


src/commands/strikes_command.py
```diff
@@ -32,12 +32,12 @@ async def add_strike(
...
         """Add a strike to the user"""
         async with create_response_context(interaction) as response_message:
             async with create_logging_embed(interaction, user=user) as logging_embed:
-                strike_details = await strike_service.get_user_strikes(
+                view_strikes = await strike_service.get_user_strikes(
                     logging_embed, user
                 )
 
-                response_message.set_string(strike_details)
+                response_message.set_string(view_strikes)
```
brtran4, line 43: we shouldn't change these two vars, just changing the function name is sufficient

brtran4 (changes requested):

src/commands/strikes_command.py
clairefleuret: done

PR #202, opened 2024-11-28
Stop exile command from being used on Mods and Admins

oytundd:
Ticket: MOD-90
/exile and /roulette(when succesfull) now errors out when they target a mod

### DIFF
--- a/src/services/exile_service.py
+++ b/src/services/exile_service.py
@@ -102,7 +102,15 @@ async def unexile_user(
             error_message,
         )
         return error_message
-
+    if user_has_role(user, Role.MOD) or user_has_role(user, Role.ADMIN):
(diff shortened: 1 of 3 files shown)


src/services/exile_service.py
```diff
@@ -33,7 +33,15 @@ async def exile_user(
...
-
+    if user_has_role(user, Role.MOD):
+        error_message = "User is a mod, no action will be taken"
+        log_info_and_add_field(
+            logging_embed,
+            logger,
+            "Error",
+            error_message,
+        )
+        return error_message
```
brtran4, line 44: let's have this mirror the check in `ban_commands.py`

brtran4 (changes requested):

brtran4 (approved):

PR #204, opened 2024-12-03
bug fix for mod-90

oytundd:
strike related functions now use interaction param when calling the exile_user() function

### DIFF
--- a/src/commands/strikes_command.py
+++ b/src/commands/strikes_command.py
@@ -22,7 +22,7 @@ def create_strikes_commands(bot: Bot) -> None:
                 interaction, user=user, reason=reason, severity=severity.name
             ) as logging_embed:
                 await strike_service.add_strike(
-                    logging_embed, user, severity, reason, interaction.user
+                    logging_embed, user, severity, reason, interaction.user, interaction
                 )
 
                 response_message.set_string(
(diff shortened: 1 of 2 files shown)


brtran4 (commented):
nit: the title of the PR should accurately describe what the code change is doing, rather than mentioning that this PR solves X ticket. that information goes in the description where we note `Ticket: MOD-XXX`

from https://devdocs.prestashop-project.org/8/contribute/contribution-guidelines/pull-requests/

> Writing a good title
> A Pull Request title should be short, but precise enough to describe the changes it introduces and how they impact the software.
> 
> Please respect the following rules:
> 
> - Be descriptive. Avoid laconic or non-descriptive sentences, like “Fix bug” or “Update file.php”.
> - Write full sentences. Avoid writing sentences without verbs, like “Yaml standards”.
> - Use the imperative mood (eg. “Add foo bar” instead of “Added…” or “Adding”). This will help keep changelog style coherent.
> - Don’t include references. Don’t add issue references like "#12314 Fix issue on…" or tags like "[BO] Fix bug where…". The PR table is there for that.
> 
> Here are some good examples:
> 
> - Fix fatal error when trying to sign up on Chrome
> - Prevent users from deleting their own profile on the users listing
> - Reduce memory usage on the Front Office home
> - Add button to close the welcome message on first login
> - Migrate product page to Symfony

PR #207, opened 2024-12-04
Add notes table to db

oytundd:
Ticket:MOD-23

Added required queries for notes table.

### DIFF
--- a/postgres/drop_all_tables.sql
+++ b/postgres/drop_all_tables.sql
@@ -3,5 +3,5 @@ begin;
 drop table IF EXISTS strikes;
 drop table IF EXISTS exiles;
 drop table IF EXISTS users;
-
+drop table IF EXISTS notes;
 commit;
--- a/postgres/drop_all_tables.sql
+++ b/postgres/drop_all_tables.sql
@@ -1,7 +1,7 @@
 begin;
-
+drop table IF EXISTS notes;
 drop table IF EXISTS strikes;
 drop table IF EXISTS exiles;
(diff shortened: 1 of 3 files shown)


postgres/drop_all_tables.sql
```diff
@@ -3,5 +3,5 @@ begin;
 drop table IF EXISTS strikes;
 drop table IF EXISTS exiles;
 drop table IF EXISTS users;
-
+drop table IF EXISTS notes;
```
KingPieson, line 6: this needs to be dropped before we drop users, because it has a FK of user ID

postgres/seed_data.sql
```diff
@@ -4,7 +4,7 @@ begin;
 delete from exiles;
 delete from strikes;
 delete from users;
-
+delete from notes;
```
KingPieson, line 7: same thing as above, you need to delete notes before you delete users

KingPieson (changes requested):

brtran4 (commented):
nit: i would title the PR for something like "Add notes table to db" or something like that

KingPieson (approved):

brtran4 (commented):
i think this should be targeted to the 1.1 hotfix branch

oytundd:
> i think this should be targeted to the 1.1 hotfix branch

i thought i would make a mod-2 branch for notes feature then work on tickets related to mod-2 under this branch then merge when its complete, is this a bad idea

PR #209, opened 2024-12-04
add tests for simple util methods

KingPieson:
Add a few simple tests for utils method. Fixed a bug with the message split method when a break happened exactly on a word, and trimmed the output

Ticket: NA


KingPieson:
ya ill do it in a follow up

PR #213, opened 2024-12-08
Create python-unittest.yml

KingPieson:
Ticket: NA

PR #214, opened 2024-12-09
Fix Role enum imports

brtran4:
Ticket: NA


brtran4:
closing - this is dealt with in #213

PR #215, opened 2024-12-10
Test utils method with Discord mocks

KingPieson:
Ticket: NA

### DIFF
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -54,32 +54,46 @@ def test_calculate_time_delta(input, expect):
 
 
 @pytest.fixture
-def create_role(mocker):
-    def __create_role(name: str):
+def create_role(mocker: MockerFixture):
+    def __create_role(name: enums.Role):
         mocked_role = mocker.Mock()
-        mocked_role.name = name
+        # name is used specifically in the Mock constructor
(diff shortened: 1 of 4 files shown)


tests/unit/test_util.py
```diff
@@ -50,3 +51,52 @@ def test_calculate_time_delta(input, expect):
...
+
+async def test_add_and_remove_role(mocker: MockerFixture, naur_guild):
+    role_to_add = enums.Role.EXILED
+    role_to_remove = enums.Role.VERIFIED
+
+    mock_add_roles = mocker.AsyncMock()
+    mock_remove_roles = mocker.AsyncMock()
+
+    # NB eventually making a mocked Member should be a fixture, but I want to understand
+    # needs a bit more before doing so
```
brtran4, line 88: im sorry these two lines are giving me a stroke

brtran4 (approved):

tests/unit/test_util.py
KingPieson: Sorry, I just wanted to call out that `Member` should be something that we can create easy mocks for. Probably we should do this via a factory method wrapped in a fixture, like I have set up for `Role` currently. I didn't put that together quite yet because I wanted to see more use cases before fully refactoring

PR #218, opened 2024-12-12
Change the strike punishment threshold to include the previous punishment if not passed

oytundd:
Ticket:MOD-137
Cumulative punishment when crossing multiple thresholds at once
Non-cumulative punishment when staying between thresholds after getting a strike

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -2,3 +2,12 @@
 MINOR_INFRACTION_POINTS = 1
 MODERATE_INFRACTION_POINTS = 3
 SERIOUS_INFRACTION_POINTS = 7
+
+# Strikes threshold and punisment array
+THRESHOLDS_PUNISHMENT = [
+    # Point threshold and punishment amount
+    (10, 14),
+    (7, 7),
+    (5, 3),
+    (3, 1),
+]
--- a/src/services/strike_service.py
+++ b/src/services/strike_service.py
@@ -49,6 +49,7 @@ async def add_strike(
 
     # increment user points, update
     db_user.last_infraction_timestamp = strike_timestamp
+    previous_points = db_user.get_strike_points()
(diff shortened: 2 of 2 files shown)


src/services/strike_service.py
```diff
@@ -120,6 +120,27 @@ def _apply_strike_point_penalty(db_user: User, severity: StrikeSeverity):
             )
 
 
+def _calculate_punishment(previous_points: int, total_points: int) -> int:
+
+    thresholds = [(10, 14), (7, 7), (5, 3), (3, 1)]
+
+    punishment_days = 0
+    speedrun = False
```
brtran4, line 128: LOL i dont really think speedrun should be the name of the variable. i think something like `jumped_multiple_thresholds`

src/services/strike_service.py
```diff
@@ -120,6 +120,27 @@ def _apply_strike_point_penalty(db_user: User, severity: StrikeSeverity):
             )
 
 
+def _calculate_punishment(previous_points: int, total_points: int) -> int:
+
+    thresholds = [(10, 14), (7, 7), (5, 3), (3, 1)]
```
brtran4, line 125: i think we should store this and comment the values in constants.py

src/constants.py
```diff
@@ -2,3 +2,12 @@
 MINOR_INFRACTION_POINTS = 1
 MODERATE_INFRACTION_POINTS = 3
 SERIOUS_INFRACTION_POINTS = 7
+
+# Strikes threshold and punisment array
+THRESHOLDS_PUNISHMENT = [
+    # Point threshold and punishment amount
```
brtran4, line 8: nit: `punishment time (in days)`

brtran4 (approved):

PR #219, opened 2024-12-16
Initial strike_service tests

oytundd:
Ticket:MOD-117


tests/unit/conftest.py
```diff
@@ -26,3 +28,55 @@ def naur_guild(mocker: MockerFixture, create_role):
...
+        mocked_user = mocker.Mock(spec=User)
+        mocked_user.user_id = user_id
+        mocked_user.discord_user_id = discord_user_id
+        mocked_user.discord_guild_id = discord_guild_id
+        mocked_user.is_mod = is_mod
+        mocked_user.temporary_points = temporary_points
+        mocked_user.permanent_points = permanent_points
+        mocked_user.last_infraction_timestamp = last_infraction_timestamp
+        mocked_user.get_strike_points = mocker.Mock(return_value=get_strike_points)
+        return mocked_user
```
KingPieson, line 54: This all can be done in the `mocker.Mock` call, the only attribute I know where that won't work is `name`

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,141 @@
...
+    [
+        (0, 1, "Nothing"),
+        (1, 8, "11 day exile"),
+        (8, 10, "14 day exile"),
+        (10, 11, "14 day exile"),
+        (1, 9, "11 day exile"),
+        (9, 12, "14 day exile"),
+        (12, 13, "14 day exile"),
+        (10, 15, "Permanent ban"),
+    ],
```
KingPieson, line 22: I think this covers a good amount of the test cases we care about, we should also cover 0 -> 3 and something that brings user past 15 points, like 10 -> 17

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,141 @@
...
+)
+async def test_apply_punisment(
+    previous_points: int,
+    total_points: int,
+    expected_punishment: str,
+    mocker: MockerFixture,
+    create_db_user,
+):
+    mocked_logging_embed = mocker.Mock()
```
KingPieson, line 31: nit: I prefer if I'm not going to be making any checks on the object to just put the mocker.Mock() call in the service call itself

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,141 @@
...
+    mocked_user = mocker.Mock(roles=[enums.Role.VERIFIED])
```
KingPieson, line 32: It doesn't matter in this case because the code you have isn't looking at the user's roles, but since this is mocking a discord version of a user, you need to pass a mocked version of the `role` object into the roles array. We already have a fixture that builds the mocked role, and I'm putting in a fixture to build the mocked user (or technically a discord member) that lets us just pass in the enum directly.

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,141 @@
...
+    mocked_db_user = create_db_user(get_strike_points=total_points)
+    res = await strike_service._apply_punishment(
```
KingPieson, line 34: Looking at `_apply_punishment`, I think you should have some patching around `ban_service.ban_user` and `exile_service.exile_user`, and you should make sure that they are being called the expected number of times and with parameters that make sense. pytest has the `mocker.ANY` that will let you match only on the exile duration if you do `assert_called_once_with` for the call to exile service

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,141 @@
+import pytest
+from pytest_mock.plugin import MockerFixture
+from moddingway.database.models import Strike
+from moddingway.services import strike_service
+from moddingway import enums
+from typing import List
+import discord
+
+
+@pytest.mark.asyncio
```
KingPieson, line 10: Are you having issues running async tests without this annotation? I think they should work without it, but if you have issues I'll make sure to add this to any async tests that I write as well

oytundd: It skips the tests when I dont have this, tells me to install plugins I already have:
 PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

oytundd: It skips the tests when I dont have this, tells me to install plugins I already have:
PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
You need to install a suitable plugin for your async framework, for example:
- anyio
- pytest-asyncio
- pytest-tornasync
- pytest-trio
- pytest-twisted
warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

KingPieson: I have the pipeline configured to install `pytest-asyncio` if you want, but I think the annotation works fine and I'll try to make sure to include it on other asyn tests I write

oytundd: It still gives the same warning after I install pytest-asyncio manually myself but im not sure what the issue actually is because google tells me putting asyncio_mode=auto in the config(which is already in there) would fix the issue.So appreciate it being included until I can find a solution.

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,178 @@
...
+
+    mocked_exile_user = mocker.patch("moddingway.services.exile_service.exile_user")
+
+    res = await strike_service._apply_punishment(
+        mocked_embed, mocked_user, mocked_db_user, previous_points
+    )
+    if total_points >= 15:
+        mocked_ban.assert_called_once_with(
+            mocked_user,
+            "Your strike were severe or frequent to be removed from NA Ultimate Raiding - FFXIV",
```
KingPieson, line 48: lets have this string be `mocker.Any`, we don't care too much about the specifics here IMO and we don't want the test to fail just because we are tweaking some user-facing messaging

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,178 @@
...
+        spy_calculate_punishment.assert_called_once_with(previous_points, total_points)
+        punishment_days = spy_calculate_punishment.spy_return
+        if punishment_days == 0:
+            mocked_exile_user.assert_not_called()
+        else:
+            mocked_exile_user.assert_called_once_with(
+                mocked_embed,
+                mocked_user,
+                mocker.ANY,
+                "Your actions were severe or frequent enough for you to receive this exile",
```
KingPieson, line 62: I don't think we need to assert on the user-facing text, this can be `mocker.Any`

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,178 @@
...
+        mocked_ban.assert_not_called()
+        spy_calculate_punishment.assert_called_once_with(previous_points, total_points)
+        punishment_days = spy_calculate_punishment.spy_return
+        if punishment_days == 0:
+            mocked_exile_user.assert_not_called()
+        else:
+            mocked_exile_user.assert_called_once_with(
+                mocked_embed,
+                mocked_user,
+                mocker.ANY,
```
KingPieson, line 61: Since this is the actual duration of the exile, we should be making sure that this matches what we expect. We should be able to pass in the datetime time duration here and have the assert work as expected, but if not we can reassess

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,178 @@
...
+    res = await strike_service.get_user_strikes(mocked_user)
+    assert res == expected_result
+
+
+@pytest.mark.asyncio
+async def test_add_strike(create_db_user, mocker: MockerFixture):
+
+    mocked_db_user = create_db_user(user_id=1, temporary_points=0, get_strike_points=0)
+
+    mocked_user = mocker.Mock(spec=discord.Member, id=1)
```
KingPieson, line 118: I merged in the PR that added the `create_member` fixture, we should be using that here. You should also update it to use the `spec` flag, and expose `id` as a field that it populates

tests/unit/test_strike_service.py
```diff
@@ -0,0 +1,178 @@
...
+
+    mocked_author = mocker.Mock(spec=discord.Member)
+    mocked_author.id = 2
+
+    mocked_logging_embed = mocker.Mock(spec=discord.Embed)
```
KingPieson, line 123: Mocked embeds might be a good candidate for a fixture as well, if you want. Not a blocker though

tests/unit/conftest.py
```diff
@@ -62,3 +69,66 @@ def __create_member(
...
+#         mocked_strike.user_id = user_id
+#         mocked_strike.severity = severity
+#         mocked_strike.reason = reason
+#         mocked_strike.created_timestamp = created_timestamp
+#         mocked_strike.created_by = created_by
+#         mocked_strike.last_edited_timestamp = last_edited_timestamp
+#         mocked_strike.last_edited_by = last_edited_by
+#         return mocked_strike
+
+#     return __create_db_strike
```
brtran4, line 134: let's either delete this or punt this to a later PR if we're going to leave this commented out

tests/unit/conftest.py
```diff
@@ -1,10 +1,13 @@
 import datetime
 from typing import List
-
 import pytest
 from pytest_mock.plugin import MockerFixture
 
+# from datetime import datetime
```
KingPieson, line 6: nit: remove unused import

KingPieson (approved):

PR #220, opened 2024-12-16
Changed a field title in strike service from Result to Strike Points 

oytundd:
Ticket:MOD-138

### DIFF
--- a/moddingway/services/strike_service.py
+++ b/moddingway/services/strike_service.py
@@ -59,7 +59,7 @@ async def add_strike(
     log_info_and_add_field(
         logging_embed,
         logger,
-        "Strike Points",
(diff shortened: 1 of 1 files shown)


src/services/strike_service.py
```diff
@@ -57,7 +57,7 @@ async def add_strike(
     log_info_and_add_field(
         logging_embed,
         logger,
-        "Result",
+        "Strike Points",
```
brtran4, line 60: i think it would be more useful to rename this as "Strike Point Adjustment" or something

brtran4 (commented):
let's point this to main. it's not urgent enough to be in the hotfix

brtran4 (approved):

PR #223, opened 2024-12-27
Allow view strikes on non-members

OTCompa:
Ticket: MOD-139
This allows moderators to use `/view_strikes` on any user not in the server with their discord id.


brtran4:
QOL question: will it grab from the cached list of users in discord and autofill in the user arg?

brtran4 (changes requested):
now that i think about it more, we should also change `view_notes` and `view_strikes` to use the same logic too

OTCompa:
> QOL question: will it grab from the cached list of users in discord and autofill in the user arg?

I don't think discord client does that, the user will have to use the user id unfortunately

brtran4 (approved):

PR #225, opened 2025-01-02
Fix issue with already-exiled users getting a strike

oytundd:
Ticket:MOD-93

### DIFF
--- a/moddingway/commands/exile_commands.py
+++ b/moddingway/commands/exile_commands.py
@@ -137,9 +137,7 @@ def create_exile_commands(bot: Bot) -> None:
     @bot.tree.command()
     @discord.app_commands.check(is_user_moderator)
     @discord.app_commands.describe(user="User whose logged exile are being viewed")
-    async def view_logged_exiles(
-        interaction: discord.Interaction, user: discord.Member
-    ):
+    async def view_logged_exiles(interaction: discord.Interaction, user: discord.User):
         """View logged exiles of a user."""
 
         async with create_response_context(interaction) as response_message:
--- a/moddingway/commands/exile_commands.py
+++ b/moddingway/commands/exile_commands.py
@@ -75,7 +75,7 @@ def create_exile_commands(bot: Bot) -> None:
                 user=user,
                 duration=format_time_string(duration),
                 reason=reason,
-                expiration=end_timestamp,
+                # expiration=end_timestamp,
             ) as logging_embed:
                 error_message = await exile_user(
                     logging_embed, user, exile_duration, reason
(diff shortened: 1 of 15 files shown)


moddingway/services/exile_service.py
```diff
@@ -28,6 +28,23 @@ async def exile_user(
...
+    db_user = users_database.get_user(user.id)
+    if db_user:
+        currentExile = exiles_database.get_user_active_exile(db_user.user_id)
+        if not user_has_role(user, Role.VERIFIED) and currentExile:
+            logger.warning(currentExile.end_timestamp)
+            new_endTimestamp = currentExile.end_timestamp + duration
+            logger.warning(new_endTimestamp)
+            exiles_database.extend_exile_duration(currentExile.exile_id, new_endTimestamp)
+            logger.info(f"Extended exile {currentExile.exile_id}")
+            log_info_and_add_field(
```
KingPieson, line 40: I think the best option to our problem here is to remove the `expiration` field from our calls to `create_logging_embed` in the commands module, and then have the service add that in the service. You'll need to configure it to happen both on the extend and on the create exile, and you should probable move the datetime -> epoch method that the embed currently is using to a utils method in some way so that we prevent duplicated code

moddingway/database/exiles_database.py
```diff
@@ -117,6 +117,22 @@ def get_user_active_exile(user_id) -> PendingExile:
             return None
 
 
+def extend_exile_duration(exile_id, end_timestamp):
```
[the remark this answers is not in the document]
KingPieson: actually after looking a bit more at what we want to do, we are going to need the end timestamp anyways for logging purposes so we should just pass in the new end date. My suggestion instead here is to just rename this method to be something like `update_exile_end`

moddingway/commands/exile_commands.py
```diff
@@ -75,7 +75,7 @@ async def exile(
                 user=user,
                 duration=format_time_string(duration),
                 reason=reason,
-                expiration=end_timestamp,
+                # expiration=end_timestamp,
```
KingPieson, line 78: nit: remove commented out code

moddingway/database/exiles_database.py
```diff
@@ -117,6 +117,23 @@ def get_user_active_exile(user_id) -> PendingExile:
...
+        UPDATE exiles
+        set endTimestamp = %s
+        where exileID = %s
+        """
+
+        params = (end_timestamp, exile_id)
+
+        cursor.execute(query, params)
+
+        return
```
KingPieson, line 134: nit: no need to explicitly return here

moddingway/services/exile_service.py
```diff
@@ -38,7 +53,7 @@ async def exile_user(
         )
         return error_message
     # look up user in DB
-    db_user = users_database.get_user(user.id)
+    # db_user = users_database.get_user(user.id)
```
KingPieson, line 56: nit: remove commented code

KingPieson (approved):

PR #229, opened 2025-01-16
Implement /delete_exile command

oytundd:
Ticket:MOD-83
Command that deletes non active(past) exiles with exile_id

### DIFF
--- a/moddingway/services/exile_service.py
+++ b/moddingway/services/exile_service.py
@@ -135,6 +135,22 @@ async def exile_user(
     )
 
 
+async def delete_exile_by_id(logging_embed: discord.Embed, exile_id: int):
+
+    result = exiles_database.delete_exile(exile_id)
+    if result:
+
+        log_info_and_add_field(
+            logging_embed, logger, "Result", f"Exile {exile_id} succesfully deleted"
+        )
+        return f"Exile {exile_id} succesfully deleted"
+    else:
+        log_info_and_add_field(
+            logging_embed, logger, "Result", f"Error with exile {exile_id} deletion"
+        )
+        return f"Error with exile {exile_id} deletion"
+
+
 async def unexile_user(
     logging_embed: discord.Embed, user: discord.Member
 ) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


moddingway/services/exile_service.py
```diff
@@ -135,6 +135,22 @@ async def exile_user(
...
+
+        log_info_and_add_field(
+            logging_embed, logger, "Result", f"Exile {exile_id} succesfully deleted"
+        )
+        return f"Exile {exile_id} succesfully deleted"
+    else:
+        log_info_and_add_field(
+            logging_embed, logger, "Result", f"Error with exile {exile_id} deletion"
+        )
+        return f"Error with exile {exile_id} deletion"
```
brtran4, line 151: we should be more specific on what this exact error is. logging error generically like this doesn't help as much when debugging

moddingway/services/exile_service.py
```diff
@@ -135,6 +135,22 @@ async def exile_user(
     )
 
 
+async def delete_exile_by_id(logging_embed: discord.Embed, exile_id: int):
+
+    result = exiles_database.delete_exile(exile_id)
+    if result:
+
```
brtran4, line 142: nit: delete this extra line

brtran4 (changes requested):

moddingway/services/exile_service.py
```diff
@@ -135,6 +135,42 @@ async def exile_user(
...
+            return f"Exile {exile_id} successfully deleted"
+        else:
+            log_info_and_add_field(
+                logging_embed,
+                logger,
+                "Result",
+                f"Error with exile {exile_id} deletion",
+            )
+            return f"Error with exile {exile_id} deletion"
+    if not status:
```
brtran4, line 167: i think since status can only be one of these cases, it would be good if we can do if/elif/elif instead of three separate ifs

brtran4 (changes requested):

brtran4 (approved):

PR #232, opened 2025-02-26
Make ban commands work on user IDs

OTCompa:
Ticket: MOD-143
Makes ban command accept both `discord.User` and `discord.Member`. Changed existing ban service from `ban_user` to `ban_member` for better clarity. Added new ban service `ban_user` which is used when a `discord.User` is given


moddingway/commands/ban_commands.py
```diff
@@ -25,22 +25,35 @@ def create_ban_commands(bot: Bot) -> None:
     )
     async def ban(
         interaction: discord.Interaction,
-        user: discord.Member,
+        user: discord.User | discord.Member,
```
brtran4, line 28: let's add a little comment on why we need to accept both a User and a Member

moddingway/services/ban_service.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger(__name__)
 
 
-async def ban_user(
+async def ban_member(
     user: discord.Member, reason: str, delete_messages: bool
 ) -> Optional[Tuple[bool, str]]:
     """Executes ban of user.
```
brtran4, line 17: let's change user to member in this comment block

moddingway/services/ban_service.py
```diff
@@ -31,29 +31,42 @@ async def ban_user(
...
+            f"Failed to ban {user.mention}. Please try again or use Discord's built-in tools.",
+        )
+
+
+async def ban_user(
+    interaction: discord.Interaction,
+    user: discord.User,
+    reason: str,
+    delete_messages: bool,
+) -> Optional[Tuple[bool, str]]:
```
brtran4, line 56: let's add a function comment similar to `ban_member`

brtran4 (commented):
can we edit the commit message on why exactly we're splitting this ban into two separate functions instead of keeping it as one? the code is a little messy for people to get it on the first read

moddingway/services/ban_service.py
```diff
@@ -31,29 +31,54 @@ async def ban_user(
...
+         interaction (discord.Interaction): Interaction that initiated this event.
+         user (discord.Member): The user being banned.
+         reason (str): Reason for the ban.
+         delete_messages_flag (bool): Whether to delete the user's messages.
+
+    Returns:
+         bool: status of the ban
+         str: any errors that occurred
+    """
+    if len(reason) >= 512:
```
KingPieson, line 69: nit, if we have a length restriction on ban_user, should we have the same on ban_member?

moddingway/commands/ban_commands.py
```diff
@@ -25,22 +25,37 @@ def create_ban_commands(bot: Bot) -> None:
...
+
         """Ban the specified user."""
         async with create_response_context(interaction) as response_message:
-            ban_success, error = await ban_user(user, reason, delete_messages)
+            if isinstance(user, discord.User):
+                ban_success, error = await ban_user(
+                    interaction, user, reason, delete_messages
+                )
+            else:
+                ban_success, error = await ban_member(user, reason, delete_messages)
```
KingPieson, line 58: Is there a benefit that we're getting for banning a member vs banning a user? I'm wondering if we can remove the condition here and just use `ban_user`, maybe only use the member-level info for seeing if the target is a mod

brtran4: iirc the objects "ban" differently, so we would have to actually split up the cases

moddingway/services/ban_service.py
brtran4: yes that is correct

brtran4 (approved):

PR #235, opened 2025-03-03
Initial fastAPI setup with example endpoint

KingPieson:
Create initial fastAPI boilerplate with an example endpoint

ticket: MOD-66

### DIFF
--- a/requirements-api.txt
+++ b/requirements-api.txt
@@ -1,2 +1,0 @@
-fastapi[standard]==0.115.11
-pydantic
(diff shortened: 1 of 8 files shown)


requirements-api.txt
brtran4, line 1: is there a reason why we can't put this in `requirements.txt`?

KingPieson: Since we're likely going to have the bot and the API be different pods, the thought was to keep the requirements separate as well. There shouldn't be any conflicts between the two, so we can keep them together if we want

brtran4: hrm i think we should go in the direction of just putting them together. having the pods have the same packages can also help us in the future if we decide to move stuff from moddingway to the api

PR #237, opened 2025-03-07
Users endpoint draft

oytundd:
Ticket:NA
early draft for users endpoint code


moddingway_api/routes/user_routes.py
```diff
@@ -3,41 +3,56 @@
...
+        User(
+            userID=str(db_user.user_id),
+            isMod=db_user.is_mod,
+            strikePoints=db_user.get_strike_points()
+        )
+        for db_user in db_user_list
+    ]
+    
+
+    return paginate(user_list)
```
KingPieson, line 58: Do you know if this properly gives us information about total record count? I would assume we either need to rely on paginate to query the DB for us, or we need to grab the total ourselves and pass it in

oytundd: image
here is what it returns after listing all the users it requested i dont know if you mean total record in the database or just the results

moddingway_api/main.py
```diff
@@ -27,3 +27,6 @@ def configure_logging():
 app = FastAPI(title="Moddingway API", lifespan=lifespan)
 
 app.include_router(user_router, tags=["user"])
+
+#this was required to run paginate() function in the route
+add_pagination(app)
```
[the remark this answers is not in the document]
oytundd: i can make a version without the pagination if you would like me to, i just assumed it was decided we are using it because i saw it being imported

moddingway_api/routes/user_routes.py
oytundd: image
yes it is just counting the results it received i will work on total records count

oytundd: implemented in my next commit

moddingway_api/main.py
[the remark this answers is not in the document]
KingPieson: IMO we should lean on the library unless it gives us issues, in which case we can set up pagination manually

KingPieson:
I'm going to spend some more time this weekend looking at the options for pagination available to us, the `paginate` method seems like the wrong approach since it expects all data to be loaded, which is unnecessary. fastapi_pagination does offer some more custom options that are still worth trying out at this point, you can take a look at them also if you'd like

oytundd:
> I'm going to spend some more time this weekend looking at the options for pagination available to us, the `paginate` method seems like the wrong approach since it expects all data to be loaded, which is unnecessary. fastapi_pagination does offer some more custom options that are still worth trying out at this point, you can take a look at them also if you'd like

 from what i understand all fastapi_pagination methods requires all data to be loaded and wants to handle pagination to be handled by database? pagination is a new concept to me so im not entirely sure but the function i set up in the database is some kind of pagination i believe

moddingway/database/users_database.py
```diff
@@ -104,3 +134,19 @@ def decrement_old_strike_points() -> int:
         cursor.execute(query)
 
         return cursor.rowcount
+    
+
+def get_user_count() -> int:
```
brtran4, line 139: this won't get the number of users in our server, it'll only get the ones that were already registered in the db (via roulette, exile, strikes, notes, etc). i'm suspecting we might have to do an api call to poll the server for the user count

KingPieson: I was under the impression that the users endpoint was just for users that we have in DB, rather than all users in the server. This is currently only being used for the users endpoint, so it doesn't need to pull from discord's API

moddingway_api/routes/mod_routes.py
```diff
@@ -0,0 +1,42 @@
...
+
+    db_user = users_database.get_user(mod_id)
+    if not db_user:
+        raise HTTPException(status_code=404, detail="User not found")
+    if db_user.is_mod:
+        mod = Mod(
+            modID = str(db_user.user_id)
+        )
+        return mod
+    elif not db_user.is_mod:
```
brtran4, line 22: else is fine here

moddingway_api/routes/mod_routes.py
```diff
@@ -0,0 +1,42 @@
...
+        raise HTTPException(status_code=404, detail="Mod not found")
+
+@router.get("")
```
brtran4, line 25: should be /mods

moddingway_api/routes/user_routes.py
```diff
@@ -1,43 +1,44 @@
...
+        raise HTTPException(status_code=404, detail="User not found")
+    user = User(
+        userID = str(db_user.user_id),
+        isMod = db_user.is_mod,
+        strikePoints = db_user.get_strike_points()
+    )
+    return user
 
 
 @router.get("")
```
brtran4, line 25: this is /users

moddingway_api/routes/mod_routes.py
oytundd: Because the route is already prefixed with /mods leaving this just an empty string already points to /mods, if I were to make this /mods we would need to request /mods/mods/

moddingway_api/routes/user_routes.py
oytundd: Same as /mods one

PR #241, opened 2025-03-16
Remove STICKY_ROLES from constants

clairefleuret:
Ticket: MOD-149
These are extra roles that should be removed from users when they are exiled. Right now there are no current roles the need to be removed other than verified. Verified is already covered in the main exile method.


brtran4 (commented):
seems like this is just https://naurffxiv.atlassian.net/browse/MOD-149?

clairefleuret:
Yeah but ptc reloaded can go too

brtran4 (approved):

PR #248, opened 2025-03-25
Assign user "Non-Verified" when joining

mmemento-morii:
Ticket: MOD-157

- Assigns user non-verified role when join

image


moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+    log_info_and_embed,
+    create_interaction_embed_context,
+    user_has_role,
+)
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+async def on_member_join(member):
```
KingPieson, line 18: These methods are pretty long and seem like they might be a good candidate for making a service and/or shared utility methods. For now it looks like you're just setting up some logs hooks so I think it's fine

mmemento-morii: Yeah I have 0 idea what I'm doing so

moddingway/bot.py
```diff
@@ -38,3 +40,9 @@ def _register_commands(self):
         create_note_commands(self)
         create_bot_errors(self)
         logger.info("Registering commands finished")
+
+    def _register_events(self):
+        logger.info("Registering event handlers")
+        self.event(on_member_join)
+        self.event(on_member_remove)
```
brtran4, line 47: i think it's more appropriate to name this as `on_member_leave`

moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+    async with create_interaction_embed_context(
+        log_channel,
+        user=member,
+        description=f"<@{member.id}> joined the server",
+        timestamp=datetime.now(timezone.utc),
+        footer="Member Joined",
+    ) as embed:
+        try:
+            # Find the role by name using the enum value
+            role = discord.utils.get(member.guild.roles, name=Role.NON_VERIFIED)
```
brtran4, line 66: since we're checking for the role in both cases, let's just move it out of the blocks, find the role, and error check beforehand

moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+                    f"Successfully assigned {role.name} role to {member.display_name}"
+                )
+            else:
+                logger.error(f"Role '{Role.NON_VERIFIED}' not found in the server.")
+        except discord.Forbidden:
+            logger.error(
+                f"Bot does not have permission to add roles to {member.display_name}"
+            )
+        except Exception as e:
+            logger.error(f"Failed to assign role: {str(e)}")
```
brtran4, line 52: i'm in favor of moving this into an error checking function

moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+        log_channel,
+        user=member,
+        description=f"<@{member.id}> left the server",
+        timestamp=datetime.now(timezone.utc),
+        footer="Member Left",
+    ) as embed:
+        try:
+            # Calculate how long the member was in the server
+            joined_at = member.joined_at
+            if joined_at:
```
brtran4, line 138: i think we can safely assume a member object will always have the `joined_at` field

moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+                time_in_server = datetime.now(timezone.utc) - joined_at
+                days_in_server = time_in_server.days
+                log_info_and_add_field(
+                    embed,
+                    logger,
+                    "Time in Server",
+                    f"{days_in_server} days (Joined: {joined_at.strftime('%Y-%m-%d')})",
+                )
+
+            # List the roles the member had
```
brtran4, line 149: this is optional, but if we want to do sticky roles for when a user leaves the server, this is the place to store this info into the database

moddingway/events/member_events.py
```diff
@@ -0,0 +1,176 @@
...
+            if member.roles:
+                role_mentions = []
+                for role in member.roles:
+                    if role.name != "@everyone":
```
brtran4, line 153: i don't think theres a role name called `everyone`?

brtran4 (changes requested):

moddingway/events/member_events.py
mmemento-morii: i think this is unnecessary as we only care about exile roles and will already be handled by the firewall

moddingway/events/member_events.py
mmemento-morii: yes but its the default role people everyone has in any disccord server, the @ is redundant i guess but thats why its there, ill clean that up

brtran4: right but it's not even going to be in `member.roles` if everyone has it. it has to be a manually assignable role

mmemento-morii: now that i think about it, how about we only check the roles we have in enums? its kinda redundant and we really dont care about clearingway roles.

image

mmemento-morii: image

moddingway/bot.py
mmemento-morii: this event in discord.py is named on_member_remove, you sure on this one? could be confusing

mmemento-morii: https://discordpy.readthedocs.io/en/latest/api.html?highlight=on_member_remove#discord.on_member_remove

moddingway/events/member_events.py
mmemento-morii: marking this as resolved, will need reviewed after rems requested changes

moddingway/events/member_events.py
brtran4: if the modding team doesnt care about ALL roles the user has, then this is fine

moddingway/util.py
```diff
@@ -194,3 +196,52 @@ def timestamp_to_epoch(timestamp: Optional[datetime]) -> Optional[int]:
...
+    Returns the channel or None if not found.
+    """
+    log_channel = guild.get_channel(settings.logging_channel_id)
+
+    if log_channel is None:
+        logger.error(
+            f"Logging channel {settings.logging_channel_id} not found. Event will not be logged to Discord."
+        )
+
+    return log_channel
```
brtran4, line 247: this is not specific to only member events, we should move this above the comment. ideally in a new PR we should probably add this check every time we are using the logging_channel_id

brtran4 (changes requested):

moddingway/bot.py
mmemento-morii: Ok, then why don't we handle both cases if a user is cached or not. If we use `on_raw_member_remove` we will lose the information that is being attached like join date, roles, etc

mmemento-morii: Actually, i think you are right. probably a good idea to use the raw event because we could have memory issues as we have a large amount of members

mmemento-morii: decided to hold off on this as we will need a singleton for when a member leaves

moddingway/events/member_events.py
```diff
@@ -0,0 +1,62 @@
...
+    # Then get the logging channel for event logging
+    log_channel = await get_log_channel(member.guild)
+    if log_channel is None:
+        return  # Exit early if no logging channel
+
+    # Create log embed
+    async with create_interaction_embed_context(
+        log_channel,
+        user=member,
+        description=f"<@{member.id}> joined the server",
```
brtran4, line 37: tbh idk if we want to keep restating the member.id because the embed gets messy after a few messages of printing the id

image

only reason why im suggesting this is because it's obvious who the targeted user is by the user field. but if we feel like we need it then im not against keeping it in

moddingway/events/member_events.py
```diff
@@ -0,0 +1,62 @@
...
+    # First find and assign the role - this happens regardless of logging ability
+    success, message, role = await find_and_assign_role(member, Role.NON_VERIFIED)
+
+    # Then get the logging channel for event logging
+    log_channel = await get_log_channel(member.guild)
+    if log_channel is None:
+        return  # Exit early if no logging channel
+
+    # Create log embed
+    async with create_interaction_embed_context(
```
brtran4, line 34: i'm scared that this will fail so we should probably wrap this in a try/except block

moddingway/util.py
```diff
@@ -194,3 +196,53 @@ def timestamp_to_epoch(timestamp: Optional[datetime]) -> Optional[int]:
     if timestamp is None:
         return None
     return round(timestamp.replace(tzinfo=timezone.utc).timestamp())
+
+
+##TODO: add this check every time we are using the logging_channel_id
```
brtran4, line 201: idk if we have a ticket for this, but this would be a good time to make a ticket and put the ticket number here

moddingway/util.py
```diff
@@ -194,3 +196,53 @@ def timestamp_to_epoch(timestamp: Optional[datetime]) -> Optional[int]:
...
+            f"Logging channel {settings.logging_channel_id} not found. Event will not be logged to Discord."
+        )
+
+    return log_channel
+
+
+# New utility functions for member events
+
+
+async def find_and_assign_role(member, role_enum):
```
brtran4, line 220: since we have so many variables this function is spitting out, it would be nice to have some return typing for this function

moddingway/events/member_events.py
```diff
@@ -0,0 +1,62 @@
...
+        description=f"<@{member.id}> joined the server",
+        timestamp=datetime.now(timezone.utc),
+        footer="Member Joined",
+    ) as embed:
+        if success:
+            log_info_and_add_field(
+                embed,
+                logger,
+                "Result",
+                f"Successfully assigned <@&{role.id}> role to {member.display_name}",
```
brtran4, line 46: we should use the `message` from `find_and_assign_role`

moddingway/events/member_events.py
```diff
@@ -0,0 +1,62 @@
...
+logger = logging.getLogger(__name__)
+
+
+async def on_member_join(member):
+    """
+    Event handler for when a new member joins the server.
+    Automatically assigns the NON_VERIFIED role and logs the action.
+    """
+    # First find and assign the role - this happens regardless of logging ability
+    success, message, role = await find_and_assign_role(member, Role.NON_VERIFIED)
```
brtran4, line 26: since `success` is a bool i would recommend to rename this to reflect that - maybe like `is_role_assigned`

brtran4 (changes requested):

moddingway/events/member_events.py
mmemento-morii: I'm confused because this is a separate pr

brtran4: it's just an example because it'll be combined together with the other PR anyways

mmemento-morii: this is the same exact format with other messages, reference the image on the op above

image

brtran4: yeah but what i'm saying idk if we want repeated @ user in the message

mmemento-morii: That should be another PR if you wanna change it because we have to go check and test every single instance an embed is being used

brtran4:
title should say "joining" not "joning"

mmemento-morii:
new embed

image

moddingway/util.py
brtran4: er sorry i meant put the ticket number in the comment

moddingway/events/member_events.py
```diff
@@ -0,0 +1,70 @@
...
+    try:
+        async with create_interaction_embed_context(
+            log_channel,
+            user=member,
+            description=f"<@{member.id}> joined the server",
+            timestamp=datetime.now(timezone.utc),
+            footer="Member Joined",
+        ) as embed:
+            if is_role_assigned:
+                # Use the message directly from find_and_assign_role which now includes the role mention
```
brtran4, line 48: this comment doesn't really describe what this block is doing. we can delete it or update it to be more accurate

brtran4 (approved):

PR #249, opened 2025-03-26
Add NON_VERIFIED role and enhance unexile_user logic

mmemento-morii:
Ticket: MOD-130
Enhance unexile_user logic
- Add NON_VERIFIED as a const
- Check if user has Non-Verified role, will not add Verified role
- Add Verified role if not already verified
- Same logic stands if already verified

### DIFF
--- a/moddingway/services/exile_service.py
+++ b/moddingway/services/exile_service.py
@@ -139,14 +139,43 @@ async def unexile_user(
     logging_embed: discord.Embed, user: discord.Member
 ) -> Optional[str]:
     if not util.user_has_role(user, Role.EXILED):
-        error_message = "User is not currently exiled, no action will be taken"
-        log_info_and_add_field(
-            logging_embed,
-            logger,
-            "Error",
-            error_message,
+        # Check if user has NON_VERIFIED role
+        if util.user_has_role(user, Role.NON_VERIFIED):
+            error_message = (
+            "User has has Non-Verified role, will not add Verified role"
         )
-        return error_message
+            log_info_and_add_field(
+                logging_embed,
+                logger,
+                "Result",
+                f"<@{user.id}> has Non-Verified role, will not add Verified role",
+            )
+            return error_message
+        
+        # Add verified role if not already verified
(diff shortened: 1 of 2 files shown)


mmemento-morii:
image
image
image

moddingway/services/exile_service.py
```diff
@@ -139,14 +139,37 @@ async def unexile_user(
...
+            return error_message
+        else:
+            error_message = "User already has Verified role, no action needed"
+            log_info_and_add_field(
+                logging_embed,
+                logger,
+                "Result",
+                f"<@{user.id}> already has Verified role, no action needed",
+            )
+            return error_message
```
brtran4, line 172: this is making the function disgustingly long, we might wanna put this in a new function

moddingway/services/exile_service.py
```diff
@@ -139,14 +139,37 @@ async def unexile_user(
...
-        log_info_and_add_field(
-            logging_embed,
-            logger,
-            "Error",
-            error_message,
-        )
-        return error_message
+        # Check if user has NON_VERIFIED role
+        if util.user_has_role(user, Role.NON_VERIFIED):
+            error_message = "User has has Non-Verified role, will not add Verified role"
```
brtran4, line 144: i wouldnt really call these error messages, the code would be acting as normal so it's more of a neutral logging message. tbh i dont even think we even need this to be user facing, but just for our own logging purposes

brtran4 (changes requested):

moddingway/services/exile_service.py
mmemento-morii: Remember, the return is the ephemeral message being given back to the user in the channel they send. It is probably good to have at least some sort of confirmation of what was done. 

Mods can these commands anywhere and would be a pain in the ass to have to to the mod reports for results

mmemento-morii: jk i see what you mean, removed error message and will do basic return

mmemento-morii:
gonna start over and close this PR since auto unexile isnt happy and need to rething this

PR #250, opened 2025-03-30
quick fix for roulette bug

KingPieson:
Ticket: MOD-168

### DIFF
--- a/moddingway/commands/helper.py
+++ b/moddingway/commands/helper.py
@@ -15,7 +15,11 @@ logger = logging.getLogger(__name__)
 
 
 def create_logging_embed(interaction: discord.Interaction, **kwargs):
-    fields = [EmbedField("Action", f"/{interaction.command.name}")]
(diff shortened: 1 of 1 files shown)


moddingway/commands/helper.py
```diff
@@ -15,7 +15,11 @@
 
 
 def create_logging_embed(interaction: discord.Interaction, **kwargs):
-    fields = [EmbedField("Action", f"/{interaction.command.name}")]
+    if interaction.command:
+        fields = [EmbedField("Action", f"/{interaction.command.name}")]
+    else:
+        # Maybe we pass something in eventually to help, for now fix crit
```
mmemento-morii, line 21: We should make a ticket and add a TODO with the ticket number so we dont forget

mmemento-morii (changes requested):

mmemento-morii (approved):

PR #252, opened 2025-04-06
Initial Users and Mods endpoints

KingPieson:
Initial work that Lemon and I put into users and mods, rolled into one PR and updated to main

Ticket: NA


oytundd (approved):

brtran4 (approved):
let's add the ticket numbers as well

PR #292, opened 2025-04-13
Refactor user retrieval logic to use find_or_create_user utility func…

memento-was-taken:


memento-was-taken:
image
image
image

moddingway/util.py
```diff
@@ -258,3 +258,46 @@ async def find_and_assign_role(
...
+
+    db_user = users_database.get_user(user_id)
+    if db_user is None:
+        if logging_embed:
+            log_info_and_embed(
+                logging_embed,
+                logger,
+                "User not found in database, creating new record",
+            )
+        db_user = users_database.add_user(user_id)
```
brtran4, line 290: we don't really need to check if the user was actually created in the db. we can trust the db to make it. let's keep only this block and return the db_user

brtran4 (changes requested):

PR #295, opened 2025-04-15
Add Glorp and Event IDs to sticky_roles

brandon-au:


moddingway/constants.py
```diff
@@ -23,4 +23,6 @@
     # Extra roles that should be removed from users when they are exiled
     # Right now there are no current roles that need to be removed other than verified
     # Verified is covered in the main method
```
memento-was-taken, line 25: Since we are here, we can remove these two lines. One is false and the other is  not needed.

moddingway/constants.py
```diff
@@ -23,4 +23,6 @@
...
+    1253810157524094996,  # Glorp
+    1272364706241187870,  # Event
```
memento-was-taken, line 27: nit: 'Event Ping' is the full role name

memento-was-taken (approved):
Looks good, just make sure to make those minor changes before you merge.

brtran4 (approved):

PR #297, opened 2025-04-20
Add event handler for member ban

KingPieson:
Ticket: MOD-269

Add DB update to create isBanned column, with default of false
add event handlers for user being banned or unbanned

### DIFF
--- a/postgres/create_tables.sql
+++ b/postgres/create_tables.sql
@@ -62,13 +62,6 @@ CREATE TABLE IF NOT EXISTS roles (
 	CONSTRAINT fk_user FOREIGN KEY(userID) REFERENCES users(userID)
 );
 
--- This can be removed after one deploy is run
--- convert all existing indefinite exiles to 1 year exiles
-UPDATE exiles SET
- exilestatus = 1,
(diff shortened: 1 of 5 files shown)


postgres/create_tables.sql
```diff
@@ -69,4 +69,9 @@ UPDATE exiles SET
  endtimestamp = starttimestamp + INTERVAL '1 year'
  WHERE exilestatus = 0;
```
brtran4, line 70: while we're here we can do some housekeeping and remove this

KingPieson: Done

postgres/create_tables.sql
```diff
@@ -62,11 +62,9 @@ CREATE TABLE IF NOT EXISTS roles (
 	CONSTRAINT fk_user FOREIGN KEY(userID) REFERENCES users(userID)
 );
 
--- This can be removed after one deploy is run
```
brtran4, line 65: i think we can still leave this here since we have to delete the ALTER query as well

KingPieson: Sounds good, readded that comment

brtran4 (approved):

Issue #265, opened 2025-04-10
Implement /unban endpoint


memento-was-taken:
/unban (2 SP)
- POST: send request to moddingway to unban user

PR #302, opened 2025-04-23
Implement unban endpoint

oytundd:
Closes:#265

### DIFF
--- a/moddingway_api/routes/banneduser_routes.py
+++ b/moddingway_api/routes/banneduser_routes.py
@@ -19,6 +19,9 @@ class BanRequest(BaseModel):
     delete_message_days: Optional[int] = 0  # optional to delete message (0-7 days)
 
 
+authHeader = {"Authorization": f"Bot {settings.discord_token}"}
+
+
 @router.get("")
 async def get_banned_users() -> Page[Banned]:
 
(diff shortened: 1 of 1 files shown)


moddingway_api/routes/banneduser_routes.py
```diff
@@ -54,3 +54,18 @@ async def ban_user(request: BanRequest):
             return {"detail": f"User {request.user_id} has been banned"}
     except (httpx.HTTPError, httpx.RequestError) as exc:
         raise HTTPException(status_code=503, detail=str(exc))
+
+
+@router.delete("")
+async def unban_user(user_id: str):
+    url = f"https://discord.com/api/v10/guilds/{settings.guild_id}/bans/{user_id}"
+    headers = {"Authorization": f"Bot {settings.discord_token}"}
```
brtran4, line 62: i think this is going to be used pretty often, let's make this a global variable at the top of the file and adjust the `ban_user` func as well

moddingway_api/routes/banneduser_routes.py
```diff
@@ -19,6 +19,9 @@ class BanRequest(BaseModel):
     delete_message_days: Optional[int] = 0  # optional to delete message (0-7 days)
 
 
+authHeader = {"Authorization": f"Bot {settings.discord_token}"}
```
brtran4, line 22: sorry this is very nitpicky, but let's put this right under router and settings (and clean up these two extra lines for min diff)

brtran4 (approved):

PR #311, opened 2025-05-19
Use get_log_channel when mod log channel accessed.

FishyBizz:


brtran4 (changes requested):
we need to retarget this to `main`, not `dev`

brtran4 (approved):

PR #312, opened 2025-05-20
Logging event bans and unbans.

FishyBizz:


FishyBizz:
Went through and rebased the two commits into one, but unsure if done correctly.

moddingway/util.py
```diff
@@ -261,3 +261,13 @@ async def find_and_assign_role(
         return False, error_msg, None
 
     # Adding to test script
+
+
+def create_logging_embed(
```
brtran4, line 266: we'll want to use `create_interaction_embed_context` instead of `create_logging_embed`

brtran4 (changes requested):
seems like theres still two separate commits. you can fix this if you do `git rebase -i` (if you used the script) or `git rebase -i origin/main` and fixup the second commit to the first

main.py
```diff
@@ -11,7 +11,11 @@
 
 if __name__ == "__main__":
     # Configure logging
-    logging.basicConfig(level=settings.log_level)
+    logging.basicConfig(
+        level=settings.log_level,
+        format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+        handlers=[logging.StreamHandler()],
+    )
```
brtran4, line 18: is there a reason why we're adding this? seems like it makes more sense in a different merge request

moddingway/events/member_events.py
```diff
@@ -84,3 +106,25 @@ async def on_member_unban(guild: Guild, user: User):
...
+
+        try:
+            async with create_interaction_embed_context(
+                log_channel,
+                user=None,
+                description=f"{user.mention} was unbanned",
+                timestamp=datetime.now(timezone.utc),
+                footer="Member Unbanned",
+            ) as embed:
+                log_info_and_add_field(embed, logger, "Action", "unban")
```
brtran4, line 124: just like line `91`, let's also preceed the action with a / to be `/unban`

brtran4 (changes requested):

main.py
FishyBizz: I had added that to see if I could change the formatting to look a little bit better and missed taking it out before my push. I will take it out and repush the updated code.

Issue #262, opened 2025-04-10
Create users and form schema


mmemento-morii:
**Users Schema:**
* Differentiate between three users:
  * **Sysadmin:**
    * Manages ban forms and active mods
    * Can deactivate mods
    * Can send out invites via moddingway
    * Discord ID
    * Role type (sysadmin)
  * **Mods:**
    * Manages ban forms
    * Discord ID
    * Role type (mod)
  * **User:**
    * Access to unban request form
    * Discord ID
    * Banned status
    * Last appeal (30 days limit until the next one)
    * Role type (user)

**Form Schema:**
* Discord ID
* Reason for appeal
* Approval
* Mod name for approval
* Timestamp

PR #315, opened 2025-05-23
Update DB user schema to handle different roles

KingPieson:
Update user db schema to allow for us to track users VS mods VS sysadmins

Ticket #262


moddingway/database/models/user.py
```diff
@@ -12,11 +13,14 @@ class User(BaseModel):
...
     temporary_points: int
     permanent_points: int
     last_infraction_timestamp: Optional[datetime] = None
     is_banned: bool
 
     def get_strike_points(self) -> int:
         return self.temporary_points + self.permanent_points
+
+    def get_is_mod(self) -> bool:
+        return self.user_role == UserRole.MOD or self.user_role == UserRole.SYSADMIN
```
brtran4, line 26: tbh i think we also probably want a separate function for sysadmin, since some actions will be only accessible to sysadmins (e.g. downloadable mod actions)

brtran4: yeah sorry i meant to say "in addition to this function".  i 100% agree with this logic we'll just need another func to check specifically for sysadmin and not mod

KingPieson: @brtran4 I updated the naming here to be `has_mod_permissions` for it to be more clear

Issue #318, opened 2025-07-10
Re-enable roulette command once bot is back online


oytundd:
roulette command disabled for temp takedown of the bot, needs to be enabled when its online

PR #323, opened 2025-07-17
enable roulette command

oytundd:
closes:#318


brtran4 (approved):
this should go in dev

PR #330, opened 2025-07-25
Update Makefile

Arctrabs:


brtran4:
a few things to note:

since this is going into `<major>.<minor>` version, we will want to point to dev for this. if the milestone was `<major>.<minor>.<patch>` then the change needed would need to go into a hotfix and will go to main. issues should always be marked with a milestone, but in the off case it's not, definitely ask me.

commit naming conventions are bit more strict than you might think (https://dev.to/shinjithdev/git-good-best-practices-for-branch-naming-and-commit-messages-oj4), the main thing to get out of this is that the title should be describing the summary of the change that starts with an imperative verb. see the commit history for some examples. personally i don't really use `<type>:` unless we really need to specify which component a PR is fixing (ex: https://github.com/naurffxiv/moddingway/pull/135). 

for this MR it could just be exactly what the ticket says: `Remove python prefix from Makefile`

brtran4 (approved):
lgtm. go ahead and merge when the pipeline checks are good

Issue #337, opened 2025-08-04
Remove sticky roles


brtran4:
## User Story
As a mod,
I want sticky roles to go away,
So that users don't see channels they're not supposed to.

## Acceptance Criteria
- [ ] Criteria 1: Users with glorp or event pings no longer have those roles when exiled

## Additional Notes
In the code itself, we shouldn't actually delete `STICKY_ROLES`, we just want to delete the content in that var

PR #339, opened 2025-08-05
remove sticky roles

oytundd:
Closes: #337


moddingway/constants.py
```diff
@@ -48,8 +48,4 @@ class UserRole(IntEnum):
     1301166606985990144: 14,  # scheduled pfs
 }
 
-STICKY_ROLES = [
-    # Extra roles that should be removed from users when they are exiled
```
brtran4, line 52: i want to leave this comment in here in case we add some other roles in the future

brtran4 (approved):

PR #345, opened 2025-08-11
Added worker startup logging and added worker keywords to log output

khadomills:
Added worker startup logging and updated log format with timestamps.

### DIFF
--- a/moddingway/workers/autounexile.py
+++ b/moddingway/workers/autounexile.py
@@ -20,8 +20,10 @@ async def autounexile_users(self):
         exiles = exiles_database.get_pending_unexiles()
     except Exception:
         logger.info("[unexile] Failed to get pending exiles.")
-        logger.info(f"[unexile] Ended auto unexile worker task with errors "
-                    f"- {datetime.now().strftime("%Y-%m-%d %H:%M:%S")}")
+        logger.info(
+            f"[unexile] Ended auto unexile worker task with errors "
+            f"- {datetime.now().strftime("%Y-%m-%d %H:%M:%S")}"
+        )
         return
 
     for exile in exiles:
@@ -31,7 +33,7 @@ async def autounexile_users(self):
             member = self.get_guild(settings.guild_id).get_member(exile.discord_id)
 
             async with create_autounexile_embed(
-                    self, member, exile.discord_id, exile.exile_id, exile.end_timestamp
+                self, member, exile.discord_id, exile.exile_id, exile.end_timestamp
             ) as autounexile_embed:
                 if member is None:
                     error_message = f"<@{exile.discord_id}> was not found in the server"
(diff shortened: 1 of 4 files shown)


moddingway/workers/autounexile.py
```diff
@@ -18,11 +19,15 @@ async def autounexile_users(self):
     try:
         exiles = exiles_database.get_pending_unexiles()
     except Exception:
-        logger.info("Failed to get pending exiles.")
+        logger.info("[unexile] Failed to get pending exiles.")
```
brtran4, line 22: so im not really sure if we need [unexile] here just because the `logger` should already have that info from whatever file it's logging from, which would be `moddingway.workers.autounexile`. so im thinking this is a little redundant. same with the other logging messages

moddingway/workers/forum_automod.py
```diff
@@ -18,14 +18,24 @@
 
 @tasks.loop(hours=24)
 async def autodelete_threads(self):
+    logger.info(
+        f"[automod] Started forum automod worker task - {datetime.now().strftime("%Y-%m-%d %H:%M:%S")}"
```
brtran4, line 22: i think we should probably include a datetime for ALL logged items, probably configurable when we first declare the `logger`

brtran4 (changes requested):

moddingway/workers/forum_automod.py
```diff
@@ -18,14 +18,18 @@
 
 @tasks.loop(hours=24)
 async def autodelete_threads(self):
+    logger.info(f"Started forum automod worker task")
+
     guild = self.get_guild(settings.guild_id)
     if guild is None:
-        logger.error("Guild not found.")
+        logger.error(" Guild not found.")
```
brtran4, line 25: nit: extra space at the beginning

moddingway/workers/forum_automod.py
```diff
@@ -18,14 +18,18 @@
...
+        logger.info(f"Ended forum automod worker task with errors")
         return
 
     notifying_channel = guild.get_channel(settings.notify_channel_id)
     if notifying_channel is None:
-        logger.error("Notifying channel not found.")
+        logger.error(" Notifying channel not found.")
```
brtran4, line 31: same nit here

moddingway/workers/strike_decrement.py
```diff
@@ -1,5 +1,7 @@
 import logging
 
+from datetime import datetime
+
```
brtran4, line 4: dont think we need this here anymore

brtran4 (changes requested):

brtran4 (approved):

moddingway/workers/autounexile.py
```diff
@@ -19,6 +19,7 @@ async def autounexile_users(self):
         exiles = exiles_database.get_pending_unexiles()
     except Exception:
         logger.info("Failed to get pending exiles.")
+        logger.info(f"Ended auto unexile worker task with errors ")
```
brtran4, line 22: Oop just caught this, nit: this file has a few trailing spaces in the logging

PR #354, opened 2025-08-12
Enhanced create_logging_embed for non-command interactions

### LINKED ISSUE #338: Roulette logging should be clearer
**Describe the bug**
The logging channel reports a `Command was run via a UI` when a user rune `/roulette` and loses

**To Reproduce**
1. Run `/roulette`
2. Lose the roulette

**Observed behavior**
The logging channel reports a `Command was run via a UI` 

**Expected behavior**
The logging channel should instead state that they lost because of a roulette

**Screenshots**

Image

**Additional context**
Proper roulette logging is actually in the code base, but it's being overridden by `Command was run via a UI`

khadomills:
Updated create_logging_embed to include optional parameter str: custom_response for logging_embed descriptions created with non-command interactions. Updated roulette command to resolve #338.


moddingway/commands/exile_commands.py
```diff
@@ -57,6 +57,7 @@ async def confirm_button_callback(
             async with create_logging_embed(
                 interaction,
                 duration=format_time_string(duration_string),
+                custom_response=f"User <@{interaction.user.id}> was exiled due to a lost roulette.",
             ) as logging_embed:
                 reason = "roulette"
```
brtran4, line 62: the original output was supposed to be this, so we'll need to clean up any dead code in this function. we had this in a `reason` field in the embed, so we'll most likely want to preserve that if possible

khadomills: Updated to pass the action and reason back into log embed. Pretty sure there's no dead code, as anything is either logged or used in the bot reply. 
image

PR #369, opened 2025-08-23
Added discord commands for manually running worker tasks

khadomills:
Added the ability for moderators to manually run auto unexile, forum automod, and strike decrement worker tasks at any time using discord slash commands.

image


moddingway/commands/forum_commands.py
```diff
@@ -0,0 +1,28 @@
...
+
+from .helper import create_logging_embed, create_response_context
+
+settings = get_settings()
+logger = logging.getLogger(__name__)
+
+
+def create_forum_commands(bot: Bot) -> None:
+    @bot.tree.command()
+    @discord.app_commands.check(is_user_moderator)
```
brtran4, line 18: nit: let's put these at the top of the function

brtran4 (commented):
i feel like these log_info_and_add_field should also log to stdout? i'm not too sure

moddingway/commands/forum_commands.py
khadomills: Verified the inner function must be decorated.

brtran4 (approved):

Issue #376, opened 2025-09-21
Moderate new event-specific party finder forum


brtran4:
## User Story
As a moderator/event staff,
I want auto moderate an event specific pf forum,
So that we can keep the forum clean during the PtC event.

## Acceptance Criteria
- [ ] Criteria 1: Add the new forum channel to the list of channels (pending @Alex-Gideon to create the forum channel)
- [ ] Criteria 2: Delete any posts older than 3 days that are sent by **ONLY** `Raid-Helper
#3806`

## Additional Notes
[Any relevant information, context, technical considerations, dependencies, or implementation details that team members should be aware of]

PR #380, opened 2025-10-04
new automod task for event forum

oytundd:
Ticket: #376 
+automod task
+automod function that checks for user
+settings for event forum
+const for event forum id and cleanup timeout
+event bot id param


oytundd:
Not entirely sure why the formatting check is failing, running formatter locally changes nothing.

brtran4:
> Not entirely sure why the formatting check is failing, running formatter locally changes nothing.

it was caused by https://github.com/naurffxiv/moddingway/commit/55a968b6813cc9b98b121ade28eed0c2d1c82469. put up a PR (https://github.com/naurffxiv/moddingway/pull/381)

brtran4 (commented):
there definitely is a way where we can code where we don't have to duplicate all of this code. the only difference between this and the auto_delete threads is that we're also checking for a user. so what we can do is if it's a certain channel (in this case the event channel), we check for the user. otherwise dont check for a user. this is a lot of thought so i'm okay with filing a tech debt for us to fix later.

can we also include the ticket number in the commit message?

moddingway/workers/forum_automod.py
```diff
@@ -20,7 +20,7 @@
 @tasks.loop(hours=24)
 async def autodelete_threads(self):
     logger.info(f"Started forum automod worker task.")
-
+    logger.info(settings.automod_inactivity)
```
brtran4, line 23: is this leftover from debugging?

moddingway/workers/__init__.py
```diff
@@ -6,4 +6,5 @@
 def start_tasks(self):
     autounexile_users.start(self)
     autodelete_threads.start(self)
+    # autodelete_event_threads.start(self)
```
brtran4, line 9: we can go ahead and delete this

brtran4 (approved):

Issue #347, opened 2025-08-11
Delete old PF posts in #pf-recruitment


brtran4:
## User Story
As a user,
I want to have a clean #pf-recruitment channel,
So that I can better see which parties are still active.

## Acceptance Criteria
- [ ] Criteria 1: Posts older than one hour should be deleted
- [ ] Criteria 2: Pinned posts should **not** be deleted

## Additional Notes
We may consider putting this in the auto forum cleanup worker OR make a new worker. I'm more in favor of sticking it into the auto forum cleanup worker

PR #399, opened 2025-10-16
Added channel cleanup worker for #pf-recruitment cleanup

khadomills:
Implemented a new scalable cleanup worker to delete posts older than a provided minutes duration. Updated constants and settings to target the pf-recruitment channel in prod. Runs every hour and logs internally, logs externally to discord channel if there's an issue removing any comments. Can be tested in a local environment with set PF_RECRUITMENT_CHANNEL_ID and DURATION variables.

Resolves Issue #347


moddingway/settings.py
```diff
@@ -47,6 +52,7 @@ def prod() -> Settings:
 
 def local() -> Settings:
     automod_inactivity = {}
+    channel_automod_inactivity = {}
```
brtran4: we should update the `README`

moddingway/constants.py
```diff
@@ -41,6 +41,7 @@ class UserRole(IntEnum):
     (3, 1),
 ]
 
+# Thread automod
 AUTOMOD_INACTIVITY = {
```
brtran4: it might be worth renaming this to `THREAD_AUTOMOD_INACTIVITY`. i'd rather have this in a separate PR

brtran4 (changes requested):

PR #404, opened 2025-10-21
Add raid time check to automod_thread helper

### LINKED ISSUE #403: Change behavior of automod task for fru event
Problem:
Automod is currently configured to delete event threads created by Raid-Helper that hast last message that is more than three days old. As a result, if a user schedules a raid that takes place more than three days in the future and no message is sent in three days, their event thread is deleted even though the raid hasn’t happened yet.

Solution:
Automod should start doing checks after the scheduled raid time, rather than three days after last message in thread.

related pr: #380

oytundd:
Introduces logic to verify if the raid time has passed in raidhelper threads before applying moderation actions. Also comments out a logger statement in member_events.py to reduce log verbosity.(Mentioned in #402)
Closes: #403

### DIFF
--- a/moddingway/events/member_events.py
+++ b/moddingway/events/member_events.py
@@ -156,7 +156,7 @@ def register_events(bot: Bot):
             if not message or not message.content:
                 logger.error(f"No message content found in event thread {thread.id}")
                 return
-            logger.info(message.content)
+            # logger.info(message.content)
             # Extract timestamp from Discord's timestamp format
             date_pattern = re.compile(r"<t:(\d+):F>")
             date_match = date_pattern.search(message.content)
--- a/moddingway/events/member_events.py
+++ b/moddingway/events/member_events.py
@@ -156,7 +156,6 @@ def register_events(bot: Bot):
             if not message or not message.content:
                 logger.error(f"No message content found in event thread {thread.id}")
                 return
-            # logger.info(message.content)
             # Extract timestamp from Discord's timestamp format
             date_pattern = re.compile(r"<t:(\d+):F>")
             date_match = date_pattern.search(message.content)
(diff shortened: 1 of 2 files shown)


moddingway/workers/helper.py
```diff
@@ -75,10 +75,32 @@ async def automod_thread(
...
     last_post = thread.last_message_id
     time_since = now - snowflake_time(last_post)
+
+    # We want to apply this check to raidhelper threads after the raid time passed
+    if (
+        user_id is not None
+        and thread.owner_id == user_id
+        and starter_message is not None
+    ):
+        date_pattern = re.compile(r"<t:(\d+):F>")
```
brtran4: should probably comment what this regex is looking for

moddingway/events/member_events.py
```diff
@@ -156,7 +156,7 @@ async def on_thread_create(thread: Thread):
             if not message or not message.content:
                 logger.error(f"No message content found in event thread {thread.id}")
                 return
-            logger.info(message.content)
+            # logger.info(message.content)
```
brtran4: if this is related to https://github.com/naurffxiv/moddingway/issues/402 we should link it in the commit message and delete this line

moddingway/workers/helper.py
```diff
@@ -75,10 +75,32 @@ async def automod_thread(
...
+            logger.error(f"No timestamp found in thread {thread.id}")
+            return num_removed, num_errors
+
+        raid_timestamp = int(date_match.group(1))
+        time_since_raid = now.timestamp() - raid_timestamp
+
+        if time_since_raid < (duration * 86400):  # 86400 seconds = 1 day
+            # logger.info(
+            #     f"Raid not started yet or duration not passed for thread {thread.id}"
+            # )
```
brtran4: let's delete this comment

brtran4 (changes requested):

moddingway/workers/helper.py
brtran4: er sorry i shouldve been more clear, what's the format of the timestamp? unix?

brtran4 (approved):

PR #405, opened 2025-10-28
refactor: remove unnecessary logging from events on prod.

### LINKED ISSUE #304: Remove unnecessary logging from event joins
logging is getting polluted from the new event joins. we should realistically only log errors instead of logging successes.

Image

obsoke:
Closes #304.

This PR changes the log level of some logging calls in the events module from `info` to `debug`. This means we should only see those calls when running moddingway locally as on prod. we should be only seeing messages with a log level of `info` or higher.

### DIFF
--- a/moddingway/events/member_events.py
+++ b/moddingway/events/member_events.py
@@ -32,7 +32,7 @@ def register_events(bot: Bot):
             member, Role.NON_VERIFIED
         )
 
-        logger.info(f"Member joined: {member.display_name} ({member.id})")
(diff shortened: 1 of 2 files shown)


obsoke:
@brtran4 I don't think I have access of any of the deployment-related environmental variables. Do you know if we are setting `MODDINGWAY_ENVIRONMENT` in prod?

brtran4 (approved):

brtran4:
all the prod settings are set in `settings.py`

PR #407, opened 2025-10-30
Removed debug print statements, added target channel to logging

### LINKED ISSUE #402: Remove debug print statements from workers\helper.py and add target channel to deletion log

khadomills:
Resolves Issue #402

### DIFF
--- a/moddingway/workers/forum_automod.py
+++ b/moddingway/workers/forum_automod.py
@@ -129,14 +129,13 @@ async def autodelete_posts(self):
 
             # Delete any messages longer than set duration
             num_removed, num_errors = await automod_channel(
-                messages,
-                duration,
+                messages, duration, channel.name
             )
 
             # Log interally if any messages are removed or failed to be removed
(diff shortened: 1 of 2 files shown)


moddingway/workers/forum_automod.py
```diff
@@ -129,14 +129,13 @@ async def autodelete_posts(self):
...
-                messages,
-                duration,
+                messages, duration, channel.name
             )
 
             # Log interally if any messages are removed or failed to be removed
             if num_removed > 0 or num_errors > 0:
                 logger.info(
-                    f"Removed a total of {num_removed} messages from channel {channel_id}. {num_errors} failed removals."
+                    f"Removed a total of {num_removed} messages from channel {channel_id} (#{channel.name}). {num_errors} failed removals."
```
brtran4: i think for readability we should probably just replace the id with the name instead. realistically theres not really much we can do with the ID

brtran4 (changes requested):

brtran4 (approved):

PR #413, opened 2025-11-04
chore: configure dependabot

### LINKED ISSUE #408: Enable Dependabot

andrewbabbittdev:
## Description

Adds the required dependabot config and pins all pip packages to specific versions so that dependabot can update them.

Fixes #408

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    open-pull-requests-limit: 30
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: ci
+
+  - package-ecosystem: "pip"
+    directories:
+      - "/**"
+    open-pull-requests-limit: 30
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: build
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    open-pull-requests-limit: 30
+    schedule:
+      interval: "daily"
```
brtran4, line 7: im kinda skeptical if daily might be too excessive since we dont commit code every single day, but we can use this as a baseline and go from there

brtran4 (approved):

.github/dependabot.yml
andrewbabbittdev: It's less about how frequently we commit code, and more about how frequently it checks for version updates to any dependencies we use.

Daily is generally the default, as Dependabot is also responsible for ensuring security updates / vulnerability checking. (You can see it under the security tab). We probably want to keep this at least weekly, if not shorter, for that reason.

brtran4: hrm yeah thats pretty important, i was thinking of every other day but if it's not commit dependent then we can leave it as daily

PR #414, opened 2025-11-04
ci: scan code with codeql

### LINKED ISSUE #410: Add CodeQL scanning
CodeQL is a code scanner from GitHub that is entirely free to use in public repos, we should create a github workflow to scan PRs and main on a frequent basis.

andrewbabbittdev:
## Description

Adds a GitHub action workflow that runs CodeQL scans.

More info on CodeQL can be found [here](https://codeql.github.com/).

Fixes #410

### DIFF
--- a/.github/workflows/scan-code.yml
+++ b/.github/workflows/scan-code.yml
@@ -1,0 +1,41 @@
+name: Scan Code
+
+on:
+  push:
+    branches: [main, dev]
+  pull_request:
+    branches: [main, dev]
+  schedule:
+    - cron: '21 9 * * 3'
+
+jobs:
+  scan-code:
+    name: Scan Code
+    runs-on: ubuntu-latest
+
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/scan-code.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4
+        with:
+          languages: ${{ matrix.language }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
```
brtran4, line 38: just a side comment but we should probably update the python-versions on the other yaml files

brtran4 (approved):

.github/workflows/scan-code.yml
andrewbabbittdev: I want to eventually get this going with `.python-version` and see if we can get it auto picked up with uv; so I'm thinking about this. We will definitely be looping back around to stuff like this 🙂

